PR #2, opened 2018-09-29
add support for automatic task provider

aurorafoss:
This pull request implements the following features:
- Detects `meson.build` and automatically create a task

Based on: [task-provider-sample](https://github.com/Microsoft/vscode-extension-samples/tree/master/task-provider-sample) from Microsoft

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+out
+node_modules
+package-lock.json
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
 	"displayName": "Meson",
 	"description": "Meson langage support for Visual Studio Code",
 	"icon": "graphics/icon.png",
-	"version": "1.1.0",
+	"version": "1.2.0",
 	"license": "Apache",
 	"publisher": "asabil",
 	"author": {
@@ -18,12 +18,34 @@
 	},
 	"homepage": "https://github.com/asabil/vscode-meson/blob/master/README.md",
 	"engines": {
-		"vscode": "^1.12.0"
+		"vscode": "^1.26.0"
 	},
 	"categories": [
 		"Programming Languages"
 	],
+	"activationEvents": [
+		"onCommand:workbench.action.tasks.runTask"
+	],
+	"main": "./out/src/extension",
 	"contributes": {
+		"taskDefinitions": [
+			{
+				"type": "meson",
+				"required": [
+					"task"
+				],
+				"properties": {
+					"task": {
+						"type": "string",
+						"description": "The Meson task to customize"
+					},
+					"file": {
+						"type": "string",
+						"description": "The Meson file that provides the task. Can be omitted."
+					}
+				}
+			}
+		],
 		"languages": [
 			{
 				"id": "meson",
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -0,0 +1,3 @@
+out
+node_modules
```
asabil: Please prefix with a / to only match the project root

package.json
```diff
@@ -1,50 +1,89 @@
 {
-    "name": "meson",
```
asabil: I might be mistaken, but package.json should always be indented with 2 spaces. Otherwise `npm install --save DEP` will rewrite the entire file changing whatever indentation to 2 spaces.

The current file is using 4 spaces, maybe we should fix this once and for all?

src/extension.ts
```diff
@@ -0,0 +1,35 @@
...
+		},
+		resolveTask(_task: vscode.Task): vscode.Task | undefined {
+			return undefined;
+		}
+	});
+}
+
+export function deactivate(): void {
+	if (mesonTaskProvider) {
+		mesonTaskProvider.dispose();
```
asabil: you might want to set `mesonTaskProvider` to `undefined` here

src/extension.ts
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { getMesonTasks } from './tasks';
+
+let mesonTaskProvider: vscode.Disposable | undefined;
+
+export function activate(_context: vscode.ExtensionContext): void {
+	let workspaceRoot = vscode.workspace.rootPath;
```
asabil: This can be a `const` instead of a `let`

src/tasks.ts
```diff
@@ -0,0 +1,83 @@
...
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { exists, exec } from './utils';
+
+interface MesonTaskDefinition extends vscode.TaskDefinition {
+	task: string;
+	file?: string;
+}
+
+export async function getMesonTasks(_dir: string): Promise<vscode.Task[]> {
```
asabil: why `_dir` not `dir` ? Typically prefixing with `_` means unused

src/extension.ts
```diff
@@ -0,0 +1,35 @@
...
+	if (!workspaceRoot) return;
+	
+	let pattern = path.join(workspaceRoot, 'meson.build');
```
asabil: Can be `const`

src/extension.ts
```diff
@@ -0,0 +1,35 @@
...
+	let mesonPromise: Thenable<vscode.Task[]> | undefined = undefined;
+	let fileWatcher = vscode.workspace.createFileSystemWatcher(pattern);
```
asabil: can be `const`

ljmf00:
Hey @asabil ,

Sorry for the typos, I'm not a javascript/typescript "fluent". As described by Microsoft on their example repo (https://github.com/Microsoft/vscode-extension-samples/blob/master/task-provider-sample/src/extension.ts#L15), they have variables instead of constants, but I completely agree with you.

src/extension.ts
ljmf00: According to [vscode documentation API](https://code.visualstudio.com/docs/extensionAPI/vscode-api#Disposable), `dispose()` actually dispose the object and free resources. It should be important to clear heap memory or something.

ljmf00:
Should be fine now :)

asabil:
Looking again at this, I am wondering why you chose to use Tasks instead of Commands?

As far as I know, you have a fixed static set of commands that are relevant to Meson.

ljmf00:
> Looking again at this, I am wondering why you chose to use Tasks instead of Commands?
> 
> As far as I know, you have a fixed static set of commands that are relevant to Meson.

Where? you mean a `Tasks` class or something? Can't you be more clear?

PR #16, opened 2020-04-09
change default build folder to builddir

stephanlachnit:
Upstream meson promotes to use `builddir` instead of `build`, see [this commit](https://github.com/mesonbuild/meson/commit/f3be687cbbfd1ff25c7ec4ffe8ba4adfee9a6579). This extension should follow that recommendation for the default build directory.


asabil:
Thank you!

PR #23, opened 2020-10-18
Add Debug command for languages that support GDB

scaryrawr:
- Based on [cmake-tools](https://github.com/microsoft/vscode-cmake-tools) debug command
- Finds the first executable defined for build and launches it under the vscode debugger
- Requires [cpptools](https://github.com/microsoft/vscode-cpptools) extension for cppdbg support
- Tested C, C++, and Vala projects with it
- Accidentally ran prettier

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,8 +8,7 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
-        "--extensionDevelopmentPath=${workspaceRoot}",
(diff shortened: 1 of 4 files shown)


Ongy:
Hey, if you are around, can you test https://github.com/mesonbuild/vscode-meson/pull/61 I used your PR as help to get the configuration stuff done (first on #57, but that knowledge helped there)

I think it supersedes this one, as it provides test configurations for every executable target.

PR #41, opened 2021-10-03
Add some basic CI

dcbaker:
At least enough to know that our typescript still transpiles


dcbaker:
This won't fire for some reason, so trying from a different branch.

PR #47, opened 2021-11-17
Fixes when there are spaces in test target name

lukester1975:
Hi

Great that this extension is now part of the meson project and will hopefully get some love.

Here are some patches to fix a few issues with spaces in test target names and to use `meson` rather than `ninja` for all commands (I use `samurai`).

I other less trivial ones but thought I'd see if these are of interest first.

Regards

### DIFF
--- a/src/meson/runners.ts
+++ b/src/meson/runners.ts
@@ -108,14 +108,9 @@ export async function runMesonBuild(buildDir: string, name?: string) {
 
 export async function runMesonTests(build: string, name?: string) {
   try {
-    if (name)
-      return await execAsTask(
-        `meson test ${name}`,
-        { cwd: build },
-        vscode.TaskRevealKind.Always
-      );
+    const args = ["test"].concat(name ?? []);
     return await execAsTask(
-      "ninja test",
+      "ninja", args,
       { cwd: build },
       vscode.TaskRevealKind.Always
     );
--- a/src/meson/runners.ts
+++ b/src/meson/runners.ts
@@ -36,7 +36,8 @@ export async function runMesonConfigure(source: string, build: string) {
           { cwd: source }
         );
         progress.report({ message: "Reconfiguring build...", increment: 60 });
-        await exec("ninja reconfigure", { cwd: build });
+        // Note "setup --reconfigure" needs to be run from the root.
+        await exec(`meson setup --reconfigure ${build}`, { cwd: source });
       } else {
         progress.report({
           message: `Configuring Meson into ${relative(source, build)}...`
@@ -110,7 +111,7 @@ export async function runMesonTests(build: string, name?: string) {
   try {
     const args = ["test"].concat(name ?? []);
     return await execAsTask(
-      "ninja", args,
+      "meson", args,
       { cwd: build },
       vscode.TaskRevealKind.Always
     );
--- a/src/meson/runners.ts
+++ b/src/meson/runners.ts
@@ -71,12 +71,15 @@ export async function runMesonReconfigure() {
 }
 
 export async function runMesonBuild(buildDir: string, name?: string) {
-  const args = name ? [name] : [];
-  const stream = execStream("ninja", args, { cwd: buildDir });
+  const title = name ? `Building target ${name}` : "Building project";
+  getOutputChannel().append(`\n${title}\n`);
+
+  const args = ["compile"].concat(name ?? []);
+  const stream = execStream("meson", args, { cwd: buildDir });
 
   return vscode.window.withProgress(
     {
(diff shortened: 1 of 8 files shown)


src/meson/runners.ts
```diff
@@ -70,12 +70,14 @@ export async function runMesonReconfigure() {
 }
 
 export async function runMesonBuild(buildDir: string, name?: string) {
-  let command = !!name ? `ninja ${name}` : "ninja";
-  const stream = execStream(command, { cwd: buildDir });
+  const title = name ? `Building target ${name}` : "Building project";
+
+  getOutputChannel().append(`\n${title}\n`);
+  const stream = execStream(`meson compile ${name ?? ""}`, { cwd: buildDir });
```
dcbaker, line 76: shouldn't this be quoted too.

dcbaker (changes requested):
Thanks for this! I have one question, but otherwise this looks good to me

src/meson/runners.ts
lukester1975: I'd only done test targets as surely nobody puts spaces in binary names 😀

That said, the one you reference ends up spawning and not going via the shell:

https://github.com/mesonbuild/vscode-meson/blob/fbede298913adb2f370466e159eb19f3ada7c842/src/utils.ts#L34-L42

So quoting breaks it there. The `FIXME` would solve the problem I suppose, though I get the feeling from looking at the code some greater refactoring to tidy up the multiple routes things can take is the right solution...

I'll update the PR for one other compile command that _is_ going via the shell for now.

lukester1975:
OK here's a revised work in progress that works with powershell. Still draft as there's a few things I need to look at, but any feedback welcome in the meantime!

The main fix was switching from `ShellExecution` -> `ProcessExecution` to just avoid any shell escaping issues. No variable expansion was required in the places I've made that change, AFAICT, so should be safe.

https://github.com/mesonbuild/vscode-meson/pull/47#discussion_r751894980 is also done now.

TODO:

1. Seemed reasonable to be consistent and use `ProcessExecution` everywhere, but there are 2 `ShellExecution`s left that I need to investigate to ensure no breakage.
2. Haven't removed `exec()` calls, which are also going via the shell. Probably a separate PR as not related to the actual issue, really.
3. Building a target in a sub-directory doesn't work on Windows... but that's nothing to do with these changes (current main branch doesn't work either). Looks to me like `meson` itself doesn't like backslashes in the target name. I'll try to take since I'm in that area.

Cheers

lukester1975:
OK here's a "final" version I'm happy with, with one caveat:

>     2. Haven't removed `exec()` calls, which are also going via the shell. Probably a separate PR as not related to the actual issue, really.

I did this in 84a3be0. But I could see an issue - if someone's used a shell variable in their configure options, then previously it would've been expanded but now won't be. Would be nice if there was an API to support vscode variable expansion and use that (e.g. https://github.com/microsoft/vscode/issues/46471) but I guess not. So maybe this commit should be binned? There's no real downside (aside from being nice to consistently avoid the shell) unless someone has spaces in their build dir, AFAICT...

>     3. Building a target in a sub-directory doesn't work on Windows... but that's nothing to do with these changes (current main branch doesn't work either). Looks to me like `meson` itself doesn't like backslashes in the target name. I'll try to take since I'm in that area.

That was a pretty trivial fix. Meson seems to be happy as long as the separator between dir and target name is a forward slash.

Hope it's useful. Makes the extension usable for me anyway!

I've got a few more other fixes but unrelated, so might as well get this in first.

dcbaker:
Would that also prevent shell variables that meson interpreters internally stop working? I do rely on that (PKG_CONFIG_PATH, in particular)

lukester1975:
Which bit in particular do you mean? The avoiding of the shell when running meson commands in general or the specific case I mention in 84a3be0?

For 84a3be0, what I'm saying is if you had say `mesonbuild.configureOptions = [ "${SOME_VARIABLE}" ]` (or in whatever format appropriate for your shell, I suppose) then there would be no expansion of `${SOME_VARIABLE}` by the shell (as there isn't one) before invoking meson. I.e. it would literally receive `${SOME_VARIABLE}` as a command line argument rather than its value.

It's the only case in the extension I could see where there might be an issue. All the other uses are just `meson <command> [--maybe a switch] maybe a target name]`.

For the first case - whatever environment VSCode has will be inherited by the meson it runs, it's just that there's no longer a shell process in the middle. So wherever meson is doing `env['PKG_CONFIG_PATH']` should get the same result.

HTH!

dcbaker:
Yeah, that should be fine. I use nix, so having environment variables is pretty important for things to work, lol

dcbaker (approved):

Issue #8, opened 2019-06-09
Element with id da39a3ee5e6b4b0d3255bfef95601890afd80709 is already registered


TheAifam5:
Hey,
When I visit the "Meson" Tab, im getting that message:
`Element with id da39a3ee5e6b4b0d3255bfef95601890afd80709 is already registered`.

meson.build:
```
project('xs', 'cpp',
  version : '0.0.1',
  meson_version : '>= 0.48.0',
  default_options : [
    'buildtype=debugoptimized',
    'warning_level=1',
    'cpp_std=c++2a',
    'default_library=both',
    'b_lundef=false',
    'b_staticpic=true',
    'b_pie=true',
    'b_lto=true'
  ])

subdir('xs_framework')
subdir('xs_tools')
```

====================================

xs_tools/meson.build:
`subdir('xs_injector')`

xs_tools/xs_injector/meson.build:
```
# Library metadata
xs_injector_name = 'xs_injector'
xs_injector_version = '0.0.1'
xs_injector_description = 'Injector'

# Define sources
xs_injector_src = [
  'src/main.cpp'
]

# Define dependencies
xs_injector_deps = [
  xs_framework_dep,
]

# Define library
xs_injector_inc = include_directories('include')
xs_injector_lib = executable(xs_injector_name,
                             xs_injector_src,
                             include_directories : xs_injector_inc,
                             dependencies : xs_injector_deps,
                             install : true)
```

====================================

xs_framework/meson.build:
```
pkg_mod = import('pkgconfig')

xs_framework_name = 'xs_framework'
xs_framework_version = '0.0.1'
xs_framework_description = 'XS Framework'

# Define c++ compile args
xs_framework_cpp_args = []

# Define sources
xs_framework_src = [
  'src/xs_framework.cpp',
]

# Define dependencies
xs_framework_deps = [
  dependency('boost', modules : [])
]

# Define library
xs_framework_inc = include_directories('include', 'pch')
xs_framework_lib = library(xs_framework_name,
                           xs_framework_src,
                           include_directories : xs_framework_inc,
                           version : xs_framework_version,
                           dependencies : xs_framework_deps,
                           cpp_args : xs_framework_cpp_args,
                           cpp_pch : 'pch/xs_pch.hpp',
                           install : true)

# Declare dependencies for this library
xs_framework_dep = declare_dependency(dependencies: xs_framework_deps,
                                      include_directories : xs_framework_inc,
                                      link_with : xs_framework_lib,
[...]

PR #48, opened 2021-11-20
Fix when tests and benchmarks have the same name

lukester1975:
Meson itself doesn't seem to mind tests and benchmarks having the same name. The extension doesn't allow it, though (or rather just fails to add benchmark nodes to the tree view due to duplicate node id).

Here's a fix and it also separates out benchmarks in to a new top level node, which I think looks nicer.

It also uses `--verbose` for running benchmarks as the output is generally of interest. But maybe that should be a config option?

And a few other tidy ups.

Cheers

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -52,12 +52,14 @@ export function activate(ctx: vscode.ExtensionContext): void {
       explorer.refresh();
     })
   );
+
   ctx.subscriptions.push(
     vscode.commands.registerCommand("mesonbuild.reconfigure", async () => {
       await runMesonReconfigure();
       explorer.refresh();
     })
   );
+
   ctx.subscriptions.push(
     vscode.commands.registerCommand(
       "mesonbuild.build",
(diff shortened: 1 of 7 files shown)


src/extension.ts
```diff
@@ -115,59 +117,63 @@ export function activate(ctx: vscode.ExtensionContext): void {
...
+      workspaceRelative(extensionConfiguration("buildFolder")),
+      isBenchmark,
+      name == "all" ? null : name
+    );
+  }
+
+  async function runTestsOrBenchmarks(isBenchmark: boolean, name?: string) {
+    name ??= await pickTest(isBenchmark);
+
+    if (name != null) {
```
dcbaker, line 156: Shouldn't this be !== ?

lukester1975: `!= null` is OK in this case as `name` will be undefined if the pick is cancelled, or "all" if the test all is selected. It could be `!== undefined`, but whilst I'm far from a Typscript guru some that I worked with preferred `!= null` when I queried the same thing. I guess it's more defensive, which seems fair.

The run build fix the other day had to be `!== null` as the code there was using null for pick cancelled and undefined for build all, so selecting all was just being treated like pick cancelled. It could probably do with some refactoring to normalise between picking a build target and picking a test :)

lukester1975:
BTW might fix #8, that's how it manifests.

src/extension.ts
lukester1975: Added another commit to the PR: aab22834b5d4cdfd9100aa5758ab5f90ecbf653b

Hopefully makes the logic a bit clearer and similar across build/test.

I notice it uses `createQuickPick` for build, presumably so it can set the busy flag whilst resolving the available build targets. 

Test/benchmarks just use `showQuickPick`. 

They should maybe be the same? Not really sure the busy flag makes sense as I think the resolution must be pretty quick as configure must have been done for it to work at all.

Indeed, if configure hasn't been done, the pick will just sit there, so that needs a bit of polish too.

dcbaker: That makes two of us who are not typescript gurus :) you're explanation makes sense to me, and I guess typescript should be able to catch the case of undefined anyway

dcbaker (approved):

PR #31, opened 2021-02-11
Rework configure prompt

thatguystone:
The configuration prompt is rather confusing and a bit annoying:

  * The dialog says "meson project detected", hinting that the dialog is only asking about this project, but "Yes" and "No" change global settings.

  * The difference between "Yes" and "This workspace" is unclear

  * If I hit "No", it continues to ask every time the project is opened, so it feels like "No" doesn't actually change anything.

The new prompt:

  * Only changes workspace settings

  * Includes "Always" and "Never" to make clear that "Yes" and "No" are one time operations

  * Allows the user to disable the prompt globally / per-workspace

PR #50, opened 2021-11-22
Refactor configure prompt

lukester1975:
Here's an update of #31 so it applies cleanly and some tidy ups.

BTW there's still some issues with the prompt in that if you open a workspace that has no meson.build and accidentally hit F1 -> Meson -> Build the extension is activated so it pops up (and then the pick target prompt just sits there). Lots of activation events other than `meson.build`...

```
  "activationEvents": [
    "onLanguage:meson",
    "onCommand:mesonbuild.configure",
    "onCommand:mesonbuild.build",
    "onCommand:mesonbuild.test",
    "onCommand:mesonbuild.benchmark",
    "workspaceContains:meson.build"
  ],
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,12 @@
             false,
             "ask"
           ],
-          "description": "Have VS Code run meson (re)configure on folder open."
+          "enumDescriptions": [
+            "Always configure on open",
+            "Never configure on open",
+            "Ask every time"
+          ],
+          "description": "Have VS Code run Meson (re)configure on opening a folder."
         },
         "mesonbuild.buildFolder": {
           "type": "string",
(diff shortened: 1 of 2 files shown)


dcbaker:
conflicts

lukester1975:
OK. Since you've merged #52 which was essentially the same thing I dunno if you're still interested in this, but I've updated it anyway.

Aside from small tidy ups to use async/await/string enum/increase DRY, I suppose it does add a couple of extra options to the dialog ("Not this time" and "Just this time") so everyone will be happy...

dcbaker:
I like the async use, and this adds some cleanliness. It'll be in the next release, thanks again!

PR #51, opened 2021-12-01
Treeview fixes

lukester1975:
I would like multiple build dir support so have worked on it - it's not quite ready, but here's some stuff I've pulled out that I think is:

* The open folder icon didn't work, so use ThemeIcon.Folder
* Some nodes used random ids and as such expansion state was lost on a refresh (expanding a nested target, hit the target to build - sub tree is collapsed). So make all node ids unique but deterministic. Also unique across vsroot+project+build dir.

Regards


Ongy (approved):
Looks good to me, an dI validated that it is a great addition to https://github.com/mesonbuild/vscode-meson/pull/53 :)

dcbaker:
Thanks for this, Sorry it's taking me so long to get back to getting this merged.

PR #52, opened 2021-12-02
Change setting mesonbuild.configureOnOpen

SelfAdjointOperator:
This commit separates the global `settings.json` and workspace `settings.json` values for `mesonbuild.configureOnOpen`. The setting now takes one of the values of `true`(bool), `false`(bool), `"ask"`(string)(default).

The previous behaviour of `false` made the extension ask the user every time they opened a workspace whether to configure or not, which was annoying. This has been changed: `false` now does not ask the user; `"ask"` will ask the user each time. `true` or `false` will run the configure task or not. This makes it easy for the user to set `mesonbuild.configureOnOpen` to `true` or `false` in their global `settings.json`, and/or have it set differently for individual projects in the workspace's `settings.json`.

The "Meson project detected" popup message for whether to configure, and set `settings.json` has been changed: Answering `"Yes"` or `"No"` will set the value of `mesonbuild.configureOnOpen` to `true` or `false` respectively in the workspace's `settings.json` (there is now no need for a `"This workspace"` option, and any answer to the popup will not affect the global `settings.json` value). Also clicking the `x` to dismiss the popup is no longer interpreted as "Yes"; dismissing does nothing as it should (this was one of the most annoying parts that needed fixing).

This commit / PR supercedes #36

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,8 +78,16 @@
       "title": "Meson build configuration",
       "properties": {
         "mesonbuild.configureOnOpen": {
-          "type": "boolean",
-          "default": false,
+          "type": [
+            "boolean",
+            "string"
+          ],
+          "default": "ask",
+          "enum": [
+            true,
+            false,
+            "ask"
+          ],
           "description": "Have VS Code run meson (re)configure on folder open."
         },
         "mesonbuild.buildFolder": {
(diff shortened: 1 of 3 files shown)


dcbaker:
Thanks for fixing this!

Issue #30, opened 2021-02-02
"Current project" sidebar is empty for project


superm1:
I've cloned https://github.com/fwupd/fwupd locally and opened it as a workspace in VS Code.  I notice that the "Current Project" sidebar has nothing in it though.  Is it supposed to show anything?

Commands like ctrl-shift-p, `Meson: configure` or `Meson: build` seem to work though.  Not sure if I'm missing out in something on that side bar?  This is the first project i'm trying to use it for.

![image](https://user-images.githubusercontent.com/494526/106631367-5304c300-6542-11eb-9595-701e37ded695.png)

adam3141:
Yes - I get the same issue. What I notice is the side bar only updates when visual studio code is restarted. If I go through a manual reconfigure, it does not update.
The following screenshot is after a Meson: Reconfigure (I changed the version from 0.1.0 to 0.1.1)
![meson-issue](https://user-images.githubusercontent.com/5292494/113791994-337b5800-9788-11eb-96b2-561fdc8fd8f1.png)
Even if I show some other side panel and go back to the Meson side panel, it does not update.

However, after a restart, this is what is displayed in the sidebar
![meson-issue-restart](https://user-images.githubusercontent.com/5292494/113792147-9240d180-9788-11eb-8c8c-707f7f72fca3.png)

Also, on an unrelated issue - you will notice it does not contain the name of my executable object which is defined in the src subdirectory.
sdlfw_exe=executable(...)

intractabilis:
What's the purpose of this sidebar? I've never seen it having more than just on line of text with the name and version of the project.

PR #53, opened 2022-02-10
fix: Reload explorer view onchanges

Ongy:
Add a filesystem watch on the "build.ninja" file in the builddir.
This allows us to reload the explorer view when anything changes.
Be it via commands from vscode or external.

I haven't tested it, but I assume that https://github.com/mesonbuild/vscode-meson/pull/51 is also very relevant as it should prevent the constant closing of tree elements


Ongy:
Forgot to mention, this is a fix for https://github.com/mesonbuild/vscode-meson/issues/30

PR #56, opened 2022-02-10
Update icon to the new meson icon

dcbaker:
@jpakkane we'd talked about this on IRC/matrix a while ago, but I'd appreciate a thumbs up/review/etc that we're not violating your copyright here.


jpakkane:
I'd say the readme already says it all so fine by me.

PR #57, opened 2022-02-11
Integrate with the native testing API (including debug)

Ongy:
While the code in this works fine, I think there's a couple of things that I'd want some feedback on first.

Mainly:
* Debug runner:
  * should we try harder to pipe the result back into tests?
  * should we try to prevent debugging non-c(++) targets? This would require some heuristics (on dependencies) I think. 
* Testing runner:
  * should we run tests in series to avoid the testlog issue?
  * If not, should we fail the tests that couldn't be run? Maybe set all to skipped, that's a state vscode supports

Other than that, just general review :) 

Commits are messy, but this grew while learning^^

closes #23

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -9,6 +9,7 @@ import {
 import { getMesonTasks } from "./tasks";
 import { MesonProjectExplorer } from "./treeview";
 import {
+  exec,
   extensionConfiguration,
   execAsTask,
   workspaceRelative,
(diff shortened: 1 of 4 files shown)


Ongy:
Git History is pretty messy, but I think the code can be merged as is.

The points I stated for the normal test runner are gone now, since I decieded to go with the simpler route to support e.g. TAP based tests as well (I need to setup gtest/tap tests and then look into doing the parallel stuff again, with maybe a patch to meson).

The debugging stuff is as stated, but I think any improvement there can go into dedicated PRs alter on. It's not broken, just missing a feature.

dcbaker (approved):

PR #60, opened 2022-02-11
refactor: move subprojects to bottom of top level node.

lukester1975:
They are currently uninteresting as you can't do anything with them so I think this is nicer.


Ongy (approved):
Should there be the same kind of filter for Tests? I thought about this but didn't want to fiddle in this code.

or should tests always be around to remind people about testing? :)

lukester1975:
Sounds reasonable to me!

Ongy (approved):

dcbaker (approved):

PR #61, opened 2022-02-11
Add debug configuration provider

### LINKED ISSUE #21: Feature request: Use dynamic debug configuration to ran targets
I think it would be beneficial for all users to have an option to run all targets via a debugger. Creating a configuration for each target is a really unpleasant job, especially if a project has 50+ tests. Thanks.

Ongy:
Planned todo: Add the build task of the debugged executable as preLaunchTask to force a rebuild.

Shares the first commit with https://github.com/mesonbuild/vscode-meson/pull/57

supersedes https://github.com/mesonbuild/vscode-meson/pull/23 IMO
Fixes #21

### DIFF
--- a/src/configprovider.ts
+++ b/src/configprovider.ts
@@ -17,15 +17,15 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
         let targets = await getMesonTargets(this.path);
 
-        let executables = targets.filter(target => target.type == "executable");
-        var ret: vscode.DebugConfiguration[] = [];
+        const executables = targets.filter(target => target.type == "executable");
+        let ret: vscode.DebugConfiguration[] = [];
 
-        for (let target of executables) {
+        for (const target of executables) {
             if (!target.target_sources.some(source => ['cpp', 'c'].includes(source.language))) {
                 continue;
             }
 
-            let targetName = await getTargetName(target)
+            const targetName = await getTargetName(target)
             ret.push({
                 type: 'cppdbg',
                 name: target.name,
--- a/src/configprovider.ts
+++ b/src/configprovider.ts
@@ -1,0 +1,37 @@
+import * as vscode from "vscode";
+
+import {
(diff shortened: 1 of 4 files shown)


Ongy:
This fixes https://github.com/mesonbuild/vscode-meson/issues/21 btw.

Ongy:
Added the mentioned `preLaunchTask`. The operation to calculate the targets became a bit more expensive. If performance becomes an issue on large projects, this might be a point for caching, and then refresh with the test controller and explorer

src/configprovider.ts
```diff
@@ -0,0 +1,50 @@
...
+        this.path = path
+    }
+
+    async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
+        let targets = await getMesonTargets(this.path);
+
+        let executables = targets.filter(target => target.type == "executable");
+        var ret: vscode.DebugConfiguration[] = [];
+
+        for (let target of executables) {
```
dcbaker, line 23: can this be `for (const target of executables) {`?

src/configprovider.ts
```diff
@@ -0,0 +1,50 @@
...
+            if (!target.target_sources.some(source => ['cpp', 'c'].includes(source.language))) {
+                continue;
+            }
+
+            let targetName = await getTargetName(target)
```
dcbaker, line 28: This can be const as well, right?

src/configprovider.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    constructor(path: string) {
+        this.path = path
+    }
+
+    async provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken): Promise<vscode.DebugConfiguration[]> {
+        let targets = await getMesonTargets(this.path);
+
+        let executables = targets.filter(target => target.type == "executable");
+        var ret: vscode.DebugConfiguration[] = [];
```
dcbaker, line 21: The two using `let` could actually be `const`, and the `var` could be `let`, right?

dcbaker (changes requested):

dcbaker (approved):
Needs a rebase, but looks good. Thanks

Issue #59, opened 2022-02-11
Sidebar broken by package.json name change


lukester1975:
Sidebar now complains about "Extension not found" since 0f4aa1cd and is completely empty. Putting the name back to "meson" instead of "mesonbuild" and all is well...

I'd have thought it would be a pretty simple fix, but after spending 10 mins on it I can't track down whatever references to "meson" might also need updating.

Anyway, I might find some time to keep looking but I thought I'd mention it in case you know the fix or want to reverse the change and do a new build in the meantime.

lukester1975:
For info, I tried a clean user and get the same problem ... so nothing to do with an old copy of extension causing a problem. Can't find anything in the log files, either. So weird.

lukester1975:
Ah! Found it. I'll open a PR.

PR #62, opened 2022-02-11
Fix extension path

lukester1975:
Per #59.

There's two commits here - first is a quick fix to change the "publisher.package" string that was broken by the package name change.

Second commit caches the extension's path from activation. Decouples from the name of the extension (and saves all those calls to getExtension()), though I don't really like having a globalish like that.

Looking at all the uses, it shouldn't be too much work to refactor to pass the extension context through to where it's needed. I don't mind looking at that if commit 2 isn't palatable, but at least the first commit should fix the problem for now!

Cheers

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -21,10 +21,13 @@ import {
   getMesonBenchmarks
 } from "./meson/introspection";
 
+export let extensionPath: string;
 let explorer: MesonProjectExplorer;
 let watcher: vscode.FileSystemWatcher;
 
 export async function activate(ctx: vscode.ExtensionContext) {
+  extensionPath = ctx.extensionPath;
+
   const root = vscode.workspace.rootPath;
   if (!root) {
     return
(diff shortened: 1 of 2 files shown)


Ongy:
I think this is fine. It's essentially a global when pulled out of the extension manager as well.

And we have worse globals in `extension.ts`, which will need to be made builddir specific once there's workspace folder/multi build config support ;)

lukester1975:
> I think this is fine. It's essentially a global when pulled out of the extension manager as well.

Except it does call `getExtension()` every time it wants the path, so feels a bit less unpleasant in that regard to me...

> And we have worse globals in `extension.ts`, which will need to be made builddir specific once there's workspace folder/multi build config support ;)

Is this something that is being worked on? I have something that I was going to polish up when I get some time, but if it's being worked on elsewhere then I guess there's no need!

Thanks

Ongy:
It's something I noticed you referenced in your PR :P

I may look into the folders, but mostly to have one workspace for meson stuff and another for javascript. Not multiple meson instances.

lukester1975:
OK cool, I'll plod on with it when I get some time then.

dcbaker:
I think this is fine in terms of "let's unbreak things" even if it isn't The Perfect™ solution. I really should have done some testing after I discovered that you can't have two extensions with the same name but different publishers (which is so strange!)

PR #67, opened 2022-02-12
build: removed array-flat-polyfill, no longer required.

lukester1975:
Small tidy up.


lukester1975:
Sorry noticed the tsconfig could be simplified, too.

PR #68, opened 2022-02-12
[RFC] Repo tidy ups

lukester1975:
Wondered if you'd like to use husky with prettier and commitlint to enforce some styling of code and commits?

I didn't like prettier at first (it doesn't format code exactly like I do!) but once I got used to letting it do its thing and not worry about it, it's just nice to have everything consistently styled. That said I don't get the predilection for 80 columns hence the custom setting. Anyway, just an idea. If you'd like to use it I can update the PR to whatever prettier/commitlint configs you like. Works OK on Windows as long as git is in the path.

3rd commit is to add standard-version for changelogs and tags if the commitlint is a go-er. I'm not sure what the current release procedure is, so might not be directly applicable and there's no single script in this, but manually `yarn release && yarn package` would do the trick.

And then one last one to have a single resources directory - couldn't see the need for separate ones but I may be missing something!

Regards


tristan957:
@dcbaker @xclaesse what are your thoughts on this? I definitely think using prettier with some config would be good if we could come up with a good format. It would help in reviews. I could go for commitlint, but I am good either way.

tristan957:
Luke, want to split the prettier stuff into another PR and this one can be just for commitlint? That way focus can be on one thing at a time?

lukester1975:
Sure I can give it a whirl when I get a moment.

tristan957:
Once the other PR is merged, do we need to add commitlint or anything else from this PR?

PR #76, opened 2022-03-25
Fix structured sources and highlight install_symlink

tristan957:


tristan957:
Can anyone think of other functions to add?

tristan957:
cc @dcbaker @xclaesse

xclaesse:
Don't we have structured doc in meson *exactly* for being able to programmatically get the list of meson functions?

tristan957:
I don't know :). This whole thing needs to be redone, but unless you know someone who is a regex genius.

xclaesse:
Could be hooked into a CI here:

```python
from pathlib import Path
import yaml

def dump(p: Path) -> None:
    if p.name.startswith('_'):
        return
    with p.open('r', encoding='utf-8') as f:
        data = yaml.load(f, Loader=yaml.CLoader)
    print(data['name'])

for p in Path('docs/yaml/builtins').glob('*.yaml'):
    dump(p)

for p in Path('docs/yaml/functions').glob('*.yaml'):
    dump(p)
```

tristan957:
You sound like you are volunteering! In all seriousness, I like your idea. But is just so easy to be lazy! @dcbaker has done cool stuff with automatic releases already as part of the CI.

dcbaker:
Can you add an entry to the change log, otherwise this looks good to me

tristan957:
Done

dcbaker:
I see the meson, and machine contestants highlighted already, or did I misunderstand?

tristan957:
I see. They aren't highlighted for me.  I am thinking the new textmate classification is better however. I can amend the PR to change the textmate classification on the original textmate object.

tristan957:
Removed my highlighting for *_target and meson. I think I need to just enable support for this in my color scheme upstream

dcbaker (approved):

PR #77, opened 2022-03-29
ci: Check all meson functions are highlighted

xclaesse:


xclaesse:
@dcbaker I don't think my CI script ran, how does this work?

xclaesse:
Functions this PR removes: {'gettext', 'find_library', 'option'}
Functions this PR adds: {'unset_variable', 'install_emptydir', 'range'}

I don't know where that gettext comes from. option is indeed needed, I think I'll just hardcode it into my script for now?

dcbaker:
Get text is super old and deprecated, just like find_library, this should definitely be removed

PR #79, opened 2022-04-01
ci: split api check into a separate job and fix indentation

dcbaker:
also clone with a depth of 1, for time savings.

I broke the CI by merging something that wasn't ready, I guess that makes it my problem to fix.


xclaesse:
LGTM

PR #80, opened 2022-04-07
Fix concatenation of multiple arguments

jfsimoneau:
When using more than one argument for setup, they all get concatenated together, making the first one the only recognized and its parameter being the concatenation of all the rest of the lines.


Ongy (approved):
Looks like a leftover from when the code went through a shell.

dcbaker (approved):

PR #87, opened 2022-05-02
Add support for muon linter and formatter

annacrombie:
A basic framework for adding further linters is also added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,21 @@
           "type": "string",
           "default": "meson",
           "description": "Specify the meson executable to use"
+        },
+        "mesonbuild.muonPath": {
+          "type": "string",
+          "default": "muon",
+          "description": "Specify the muon executable to use"
+        },
+        "mesonbuild.linting.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Globally enable/disable linting"
+        },
+        "mesonbuild.linter.muon.enabled": {
+          "type": "boolean",
+          "default": false,
+          "description": "Enable muon linter"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -128,6 +128,22 @@
           "type": "boolean",
           "default": false,
           "description": "Enable muon linter"
+        },
+        "mesonbuild.formatting.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Globally enable/disable formatting"
+        },
+        "mesonbuild.formatting.provider": {
+          "type": "string",
+          "default": "muon",
+          "enum": ["muon"],
+          "description": "Select which formatting provider to use"
+        },
+        "mesonbuild.formatting.muonConfig": {
+          "type": "string",
+          "default": null,
+          "description": "Path to muon formatter config.ini"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
         },
         "mesonbuild.formatting.enabled": {
           "type": "boolean",
-          "default": true,
+          "default": false,
           "description": "Globally enable/disable formatting"
         },
         "mesonbuild.formatting.provider": {
--- a/src/linters.ts
+++ b/src/linters.ts
@@ -34,6 +34,8 @@ async function reloadLinters(sourceRoot: string, context: vscode.ExtensionContex
     return disposables;
   }
 
+  let enabled_linters: ((document: vscode.TextDocument) => Promise<vscode.Diagnostic[]>)[] = [];
+
   for (const [name, props] of Object.entries(linters)) {
     const config: LinterConfiguration = extensionConfiguration("linter")[name];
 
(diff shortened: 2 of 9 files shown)


Ongy:
Cool stuff :)
I'm not sure if I like findExecutable code. Is there a reason you did it this way? For `mesonPath` we just use the path resolution and have it default to `meson` to avoid doing this kind of handling by hand in the execution.


If we decide that we want to have muon as supported tool in this extension, it would probably make sense to also wire in the formatter.

src/linters.ts
```diff
@@ -0,0 +1,164 @@
...
+      continue;
+    }
+
+    let linter_path = config.path;
+    if (!linter_path) {
+      linter_path = await findExecutable(name);
+    }
+
+    if (!linter_path) {
+      continue;
```
Ongy, line 128: If it's enabled and cannot be found we should likely create an error message (notification/popup) about that.

annacrombie:
The point of `findExecutable` is so that in the future, when muon analyze is more stabilized/tested, it can be enabled by default but won't cause problems if the executable is not found.

I only disabled it by default for now since it is still somewhat experimental, and there are still a few confusing false positives.

I agree about adding the formatter, I will add that to this PR later today hopefully.  I was wondering how vscode extensions usually handle tools with external configuration.  muon fmt can take a .ini file to configure its behavior, do extensions usually map external options to vscode configuration options?  Or would a `path_to_config` option be sufficient?

package.json
```diff
@@ -113,6 +113,16 @@
           "type": "string",
           "default": "meson",
           "description": "Specify the meson executable to use"
+        },
+        "mesonbuild.linters.muon.enabled": {
```
tristan957, line 117: maybe `mesonbuild.linter` is a better option name with Muon being the default? Presumably more linters could be added in the future.

Alongside this option could be `mesonbuild.linting.enabled` defaulting to true.

package.json
```diff
@@ -113,6 +113,16 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "enable muon linter"
```
tristan957, line 120: Capitalize E in enable

package.json
```diff
@@ -113,6 +113,16 @@
...
+        },
+        "mesonbuild.linters.muon.path": {
+          "type": "string",
+          "default": null,
+          "description": "custom path for muon executable"
```
tristan957, line 125: capitalize C in custom

Ongy:
I think path to config is fine.
Maybe have some way to default to something next to the toplevel meson.build (if that's not already in muon ;) )

Settings can be global / workspace / folder. So we get a hierarchy and flexibilty from the vscode side.


Is the `findExecutable` much better than trying to execute it (to read version and determine if good?) and use that?

package.json
```diff
@@ -113,6 +113,36 @@
...
+          "type": "boolean",
+          "default": false,
+          "description": "Enable muon linter"
+        },
+        "mesonbuild.formatting.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Globally enable/disable formatting"
+        },
+        "mesonbuild.formatter.muon.enabled": {
```
tristan957, line 137: I don't these need to exist anymore, the `*.muon.enabled` settings. We should just have `mesonbuild.linter` which user can set to `muon` and `mesonbuild.formatter` which user can also set to `muon`. Those should be the defaults.

annacrombie: The linting framework was designed to support multiple linters (like [ale](https://github.com/dense-analysis/ale)) simultaneously.

The formatting framework is basically the same as the linting one so it also supports multiple formatters simultaneously, though I'm not sure what vscode does if it has multiple registered formatters.

tristan957: You can pick a default formatter if more than one exists.

annacrombie: So given what I said, do you still think the options need to be changed?

annacrombie: Alright, I kept the formatting config the same, but I changed the formatting config to have a provider.  I think I like having enabled / provider separate rather than having `"none"` be a provider so I left it that way for now.

tristan957:
Does Muon also accept formatting options as command line parameters? I am not familiar enough to know.

annacrombie:
@tristan957 no, muon fmt doesn't accept any formatting configuration options on the command line, just an optional path to a config.ini

tristan957:
Then this looks pretty good to me! I don't remember if I saw it in this PR, but does the extension have the ability to download a pre-compiled muon if it doesn't exist on the user's system?

tristan957:
Maybe I am just getting ahead of myself :). Can always add muon as a subproject and use it from your builddir :)

annacrombie:
Awesome!  No, this extension won't download a muon binary for you.  I feel like that is a little too assuming for my taste.  There are [static binaries](https://muon.build/releases/) available to download though, and the linting / formatting stuff should just work if you download the binary, make it executable, and point `muonPath` at it.

tristan957:
Yea I was thinking that maybe this could be like the clangd extension for instance, where the latest and greatest can be downloaded through the extension if you don't have it.

dcbaker (approved):
LGTM.

Thanks for working on this, and thanks everyone for reviewing it :)

dcbaker:
@ongy, are you happy with this now?

annacrombie:
@dcbaker, I rebased on main but it seems like the CI didn't re-trigger when I force-pushed?

dcbaker:
@annacrombie because you've never commited anything to the repo github won't let you auto-trigger CI to make sure you're not some kind of coin mining bot… crypto can't die fast enough

annacrombie:
:tada:

PR #99, opened 2022-08-20
update tool version handling

annacrombie:
Pass around a `Tool` that consists of a version + a path, so that the linting / formatting functions can alter their behavior based on version.  Also, make the version check for muon smarter.

### DIFF
--- a/src/formatters.ts
+++ b/src/formatters.ts
@@ -5,12 +5,13 @@ import {
   getOutputChannel
 } from './utils';
 import {
-  ToolCheckFunc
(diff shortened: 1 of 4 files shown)


Ongy (approved):

Ongy:
Not the biggest fan to do this at all, but the underlying tooling requires it.
And while I'd prefer a capability auto-detection, parsing versions is more sane and stable than trying to parse help output...

annacrombie:
Well, the capabilities will reliable depending on the version, that's the whole point of semver.

PR #103, opened 2022-12-23
WIP: multiple workspace and build dirs

lukester1975:
I mentioned some time ago I had some hacks to support multiple workspace and build dirs. I've never really found the time to finish it properly, just have the minimum that I required for my usage.

R.e. build dirs, the logic is to use the existing config option as the default build directory (in picker dialogs etc.) but to walk the parent of that directory for other meson build dirs. I personally set `builds/debug` as my default builddir, such that `builds` is checked rather than a project's root (but a depth of 1 should be fine). 

Anyway, I have tidied up what I have and rebased on current main so thought I'd mention it here in case it's of interest. I may spend the odd hour on it every now and then, so it might progress a bit further. But most of the time I just feel a clean sheet of paper would be better!

Fixes in a203a96d547cb4c618cdf6ec1e6b76313174985a, 49d4d67e3dcde8a8e1b36f2d4a25e945b6a01789 would be nice to get merged in. c9d3c33774ac711ce7934a7895a1810d42735c48 too, though that's my personal preference and maybe it should be a config option? Happy to do that if so.

Cheers 

---

* Moved configure button to project node, for multiple root workspace support.

* Made runMesonBuild() go through vscode.Task interface.
  * This affects building via clicking on a target in the meson treeview, or
    running the "Meson: Build" command.
  * "Progress" pop up is gone.
  * Terminal output is now is revealed and copes with escape sequences.

* Don't display "Current project" in meson sidebar. Confusing with multiple.

* Added context menu to builddir, target and test nodes in sidebar.

TODO

* remove all uses of vscode.workspace.root
* getTargetName(). Might not be needed at all? meson doesn't need path to target name now?
  Otherwise fix up wrt relevant workspace folder.
* All workspaceRelative(), getBuildFolder() usage.
* Maybe the concept of "active build directory" is nicer to clean up command palette usage (too many pick dialogs when there are multiple roots and build dirs within them).


tristan957:
@lukester1975 will you be coming back to this?

lukester1975:
Well, whilst I am having a few days chasing/cleaning up various issues, last time I tried rebasing this after the big changes in 1.9 it made my head hurt too much! As such, probably a longer job than I can give time to right now. Possibly easier to just start again with it as a guide... which I would like to do, but not sure when.

Unless meson has a Monster fund and wants to send me a crate or ten 😀

tristan957:
> Well, whilst I am having a few days chasing/cleaning up various issues, last time I tried rebasing this after the big changes in 1.9 it made my head hurt too much! As such, probably a longer job than I can give time to right now. Possibly easier to just start again with it as a guide... which I would like to do, but not sure when.
> 
> Unless meson has a Monster fund and wants to send me a crate or ten 😀

If only! No rush. Do you think we should close this then?

lukester1975:
Sure, fine by me.

tristan957:
Cool thanks for your contributions to the project thus far.

PR #106, opened 2023-01-13
Do not reconfigure on vscode startup

xclaesse:
Reconfiguring a project can be slow, especially when having lots of subprojects. Once the project has been configured, there is no reason to reconfigure it on startup, a rebuild will reconfigure only if needed.

### DIFF
--- a/src/meson/runners.ts
+++ b/src/meson/runners.ts
@@ -28,22 +28,7 @@ export async function runMesonConfigure(source: string, build: string) {
       const configureOpts = extensionConfiguration("configureOptions");
       const setupOpts = extensionConfiguration("setupOptions");
 
-      if (await checkMesonIsConfigured(build)) {
-        progress.report({
-          message: "Applying configure options...",
-          increment: 30
-        });
-
-        await exec(
-          extensionConfiguration("mesonPath"), ["configure", ...configureOpts, build],
-          { cwd: source }
-        );
-        progress.report({ message: "Reconfiguring build...", increment: 60 });
-
-        // Note "setup --reconfigure" needs to be run from the root.
-        await exec(extensionConfiguration("mesonPath"), ["setup", "--reconfigure", ...setupOpts, build],
-          { cwd: source });
-      } else {
+      if (!await checkMesonIsConfigured(build)) {
         progress.report({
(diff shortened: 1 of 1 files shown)


xclaesse:
I'm new to vscode extensions, how can I test local modifications to an extension?

tristan957:
@xclaesse open the repo in vscode, and then launch a debug session.

xclaesse:
@tristan957 I get that error:
```
Activating extension 'mesonbuild.mesonbuild' failed: Cannot find module '/home/xclaesse/programmation/vscode-meson/out/src/extension' Require stack: - /usr/share/code/resources/app/out/vs/loader.js - /usr/share/code/resources/app/out/bootstrap-amd.js - /usr/share/code/resources/app/out/bootstrap-fork.js.
```

tristan957:
Seems like the extension wasn't built.

xclaesse:
@dcbaker @tristan957 ok, tested this PR and it works fine. What do you think?

tristan957 (approved):

ylatuya:
I am not sure if this fix is correct. There is already an option to configure that behaviour `configureOnOpen` and the user is asked the first time to choose between the different options already: https://github.com/mesonbuild/vscode-meson/blob/main/package.json#L96

xclaesse:
@ylatuya I want to configure on open when builddir does not exist, but not reconfigure for no reason when builddir already exists.

ylatuya:
@xclaesse I don't think this matches the existing description and should probably be a new option:

```
          "enumDescriptions": [
            "Always configure on open",
            "Never configure on open",
            "Ask every time"
          ],
          "description": "Have VS Code run Meson (re)configure on opening a folder."
```

If I select  "Always configure on open", I would expect from the description to reconfigure on opening if already reconfigured. Whether that makes sense or not, it's the way it's documented right now.

xclaesse:
Should definitely remove the `(re)` in description, what else would you suggest?

ylatuya:
I would also rename  `"Always configure on open"` to  `"Automatically configure on open"` in the enum description. With that it's clear that configuration is automatic, not always. CMake's extension uses the same wording.

PR #108, opened 2023-01-19
Fix description for configureOnOpen

ylatuya:
Change the description to match new behaviour
from https://github.com/mesonbuild/vscode-meson/pull/106


ylatuya:
cc: @xclaesse

xclaesse:
LGTM, but I don't have push permission. @dcbaker

PR #110, opened 2023-01-20
Fix automatic executable targets execution 

ylatuya:
* Add support for LLDB and VS debuggers
* Fix targets  pre launch compilation task

Depends on https://github.com/mesonbuild/vscode-meson/pull/109

### DIFF
--- a/src/configprovider.ts
+++ b/src/configprovider.ts
@@ -1,18 +1,66 @@
 import * as vscode from "vscode";
-
+import * as nls from 'vscode-nls';
+import * as path from 'path';
 import {
   getMesonTargets
 } from "./meson/introspection"
 import {
+  Target
+} from "./meson/types"
+import {
   extensionConfiguration,
   getTargetName
 } from "./utils"
+
+nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
+const localize: nls.LocalizeFunc = nls.loadMessageBundle();
+
+export enum MIModes {
+  lldb = 'lldb',
+  gdb = 'gdb',
+}
 
 export class DebugConfigurationProvider implements vscode.DebugConfigurationProvider {
   private path: string;
 
   constructor(path: string) {
     this.path = path
+  }
+
+  async createBaseDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
+    const targetName = await getTargetName(target);
+    return {
+      type: 'cppdbg',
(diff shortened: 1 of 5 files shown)


src/configprovider.ts
```diff
@@ -1,54 +1,102 @@
...
+      args: [],
+      preLaunchTask: `Meson: Build ${targetName}`
+    };
+  }
 
-        let configDebugOptions = extensionConfiguration("debugOptions");
+  async createGDBDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
+    let debugConfig = await this.createBaseDebugConfiguration(target);
+    debugConfig.MIMode = MIModes.gdb;
+    debugConfig.setupCommands = [{
```
Ongy, line 46: I'm mildly convinced that I removed that on purpose because some doc says it may be an issue.

And there is a default config set that should merge in some setupCommands
Can you check the options and whether the startupCommands or debug thingy in there is considered for this?

tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "target": "es6",
+    "target": "es2016",
```
Ongy, line 4: just for confirmation, are these updates necessary for the new package?

Otherwise I think they shouldn't be in this change (but are generally appreciated).

Ongy (commented):
While I haven't used it myself, My understanding is that github has a feature of "stacked PRs" https://blog.logrocket.com/using-stacked-pull-requests-in-github/

If it's possible without too much trouble, could you split the cleanup commit and stack on that?
Makes review more sane.
to be clear: I don't consider this a merge blocker

tsconfig.json
ylatuya: The linter was suggesting the upgrade with some API I started using  and I upgraded it without checking it, but it's not really needed. I will remove it

ylatuya:
> and

It was already done that way, I created a PR with that commit here: https://github.com/mesonbuild/vscode-meson/pull/109
which has been merged already

Ongy (approved):

src/configprovider.ts
ylatuya: I tried replicating the same configuration used by CMake's extension https://github.com/microsoft/vscode-cmake-tools/blob/896f499b99a85e60b36bd8b8bf5a5e779af14236/src/debugger.ts#L94.

I don't know how to check if there is any config already provided, to be honest. I tried getting it from the context without luck.

PR #113, opened 2023-01-27
Show configure errors

lukester1975:
Kinda important! Without, a modal dialog:

`Command Meson: Configure resulted in an error (An unknown error occurred. Please consult the log
for more details.)`

Not very helpful. With, the meson build section of the output pane will contain the failed configure command's output.

Also show a bit more info in the final progress dialog update (already configured, configured ok,
failed).

And a couple of fixes in separate commits.


xclaesse:
With https://github.com/mesonbuild/vscode-meson/pull/127 the configuration runs into a terminal, so errors are more obvious.

xclaesse:
Configure is now run in terminal, so I think this can be closed.

PR #118, opened 2023-03-31
Remove ccache from compile_commands.json

xclaesse:
It confuses vscode's intellisense, causing it to not understand GNUC
    syntaxes for example. The bug was reported upstream:
    https://github.com/microsoft/vscode-cpptools/issues/7616


dcbaker:
can you add a CHANGELOG entry for the ccache removal? That seems like it's worth documenting.

xclaesse:
> can you add a CHANGELOG entry for the ccache removal? That seems like it's worth documenting.

done

src/utils.ts
```diff
@@ -166,3 +166,27 @@ export async function genEnvFile(buildDir: string) {
...
+    var cmd = db[i]['command'].split(" ");
+    if (cmd[0].endsWith("ccache")) {
+      cmd.shift();
+      db[i]["command"] = cmd.join(" ");
+      changed = true;
+    }
+  }
+  if (changed) {
+    fs.writeFileSync(filePath, JSON.stringify(db, null, "  "));
+  }
```
dcbaker, line 191: you may be intesrsted in `for (let i of db)`, which is equivalent to python's `dict.values()`. Alternatively, I think this could be rewritten as:
```typescript
let changed = false;
Object.values(db).forEach((entry) => {
  let cmd = entry.command.split(" ");
  if (cmd[0].endsWith("ccache")) {
    cmd.shift();
    entry.command = cmd.join(" ");
    changed = true;
});
```

using these functional patterns is more idiomatic for javascript/typescript than writing for loops.

Also, in javascript/typescript `foo.bar` is the same as `foo["bar"]`. I haven't decided yet whether this is terrifying or amazing

src/utils.ts
```diff
@@ -166,3 +166,27 @@ export async function genEnvFile(buildDir: string) {
...
+  try {
+    var filePath = path.join(buildDir, "compile_commands.json")
+    var db = JSON.parse(fs.readFileSync(filePath, "utf-8"));
+  } catch {
+    return;
+  }
+  var changed = false;
+  for (var i in db) {
+    // FIXME: This should use proper shlex.split() and shlex.join()
+    var cmd = db[i]['command'].split(" ");
```
dcbaker, line 182: in these cases you should use `let` instead of `var`. `var` has some… interesting… behavior, which you're taking advantage of in the try/catch block

src/utils.ts
```diff
@@ -166,3 +166,27 @@ export async function genEnvFile(buildDir: string) {
...
+
+export async function patchCompileCommands(buildDir: string) {
+  // Remove ccache from compile commands because they confuse Intellisense:
+  // https://github.com/microsoft/vscode-cpptools/issues/7616
+  try {
+    var filePath = path.join(buildDir, "compile_commands.json")
+    var db = JSON.parse(fs.readFileSync(filePath, "utf-8"));
+  } catch {
+    return;
+  }
```
dcbaker, line 178: I don't think `path.join` is going to throw here, it just returns a string like python's `os.path.join()`.

`JSON.parse()` will throw a syntax error if the json is invalid, which seems unlikely, but it's possible.

I think we might be better off using the asynchronous read function though, since compile_commands.json can be quite large.

So I think that would look something like:
```typescript
const filePath = path.join(buildDir, "compile_commands.json");
let db = await fs.readFile(filePath, "utf-8", (err, data): Error | object => {
  if (err) { return err; }
  try {
    return JSON.parse(data);  /// might need to add `as object` here...
  } catch (e) {
    return e;
  }
});

if (db instanceof Error) { return; }
```

dcbaker (changes requested):

src/utils.ts
xclaesse: > Object.values(db).forEach((entry) => {

did that

> Also, in javascript/typescript foo.bar is the same as foo["bar"]

I think it needs to be propertly typed for that, at least the syntax highlighter is not happy with `entry.command`.

src/utils.ts
xclaesse: Even better, we already had parseJSONFileIfExists() for that.

src/utils.ts
xclaesse: var makes it available outside of the scope, like python, right?

xclaesse:
@dcbaker I changed a bit how this works, it now generates a separate file instead of patching existing compile_commands.json. That's better in case something else is using that file. Also avoids potential issue of recursive file changed notification because we were writing a file in the watcher callback of that file.

src/utils.ts
dcbaker: `var` does what's called "hoisting" basically anything declared with `var` is equivalent to being called with `let` at the very top of the function, uninitialized. so you can write something intresting like
```javascript
() => {
  console.log(foo);  // undefined, but exists
  var foo = 'foo';
}
```
because this is equivalent to
```javascript
() => {
  var foo;
  console.log(foo);
  foo = 'foo';
}
```

because of the hoisting, you also end up with var variables having function rather than block scope, so you can do such silliness as:
```javascript
const func = () => {
  let foo = () => { 
    var bar = 7;
    return bar
  };
  bar = 8;
  return foo;
}
func();  // returns 8
```

dcbaker (approved):

Issue #119, opened 2023-03-31
[Feature] Integrate language server 


JCWasmx86:
I wrote a language server for meson: https://github.com/JCWasmx86/Swift-MesonLSP (GPLv3, but currently trivially changeable as I'm the only contributor)

This is a tracking issue for integration in this extension.

## Requirements
- [x] No licensing issues
- [x] Works on Windows (A bit untested compared to Linux/MacOS)
- [x] Works on Linux
- [x] Works on MacOS
- [x] Binaries are attached to releases

## Open questions
- How to ship it using distro mechanisms on Linux? NixOS seems to have support for building Swift packages, for Fedora I could make use of COPR

I made a fork of vscode-meson for testing: https://github.com/JCWasmx86/vscode-meson It's quite barebones, it just starts the language server. Advanced options/Download of binaries are not implemented due to not being that well versed in writing VSCode extensions

dcbaker:
I'll write a nix expression so we can test the distro packaged aproach

JCWasmx86:
I've setup a COPR for Fedora 37, here: https://copr.fedorainfracloud.org/coprs/jcwasmx86/Swift-MesonLSP/

JCWasmx86:
An APT Repo for Ubuntu 18.04/20.04/22.04 and Debian Stable/Testing/Unstable is set up here: https://github.com/JCWasmx86/swift-mesonlsp-apt-repo

tristan957:
Perhaps this document should just be part of the extension. I don't really see why it couldn't.

I think we should stand on the shoulders of giants in this regard. I haven't looked at how it works, but the clangd extension has pretty good update functionality.

JCWasmx86:
Yes, makes sense. But if the JSON file is in the extension, it couples the language server to the extension version. But I will implement it in some next PR (Expect it sometime in the next five weeks)

tristan957:
I don't really think that is a big deal if I am being honest. Maybe @dcbaker has a different opinion.

siberianbot:
If I understood correctly, mesonlsp have `--path` parameter, which defines which file to parse as root. It will be very useful to specify path to root `meson.build` through VS Code settings.

I.e. my monorepo setup have several projects with different languages/build systems involved. mesonlsp looks for meson.build in root directory of workspace and I see such error messages in log:
```
[ WARN ] analyze::mesontree - ../src/libanalyze/mesontree.cpp:195: No meson.build file in {root directory name here}
```

[...]

PR #123, opened 2023-04-06
Add integration for Swift-MesonLSP

JCWasmx86:
See #119

### Features
- [x] Download and setup binaries for MacOS/Windows (For windows: Requires a zip file due to depending on some DLLs that can't be statically linked)
  - [x] Download binaries
  - [x] Check checksum
  - [x] Extract zip file (Requires external library)
  - [x] Make lsp.ts use the downloaded file
- [x] Code style/Formatting
- [x] Figure out if something went wrong with yarn.lock and if yes what went wrong (Doesn't seem to be something bad)
- [x] Get rid of curl usage to download file
- [x] Don't require users to restart VSCode 
- [ ] Add language server to https://langserver.org/ ? (See https://github.com/JCWasmx86/Swift-MesonLSP/issues/11)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## next
 
+- Add integration with
+  [Swift-MesonLSP](https://github.com/JCWasmx86/Swift-MesonLSP). vscode-meson
+  will automatically download it on Windows, Linux and MacOS. Linux users can
+  follow the [docs](https://github.com/JCWasmx86/Swift-MesonLSP/tree/main/Docs)
+  if they want to have updates delivered using their package manager.
+
 ## 1.11.0
 
 - Add syntax highlight for `meson.options` files.
--- a/package.json
+++ b/package.json
@@ -86,6 +86,10 @@
       {
         "command": "mesonbuild.benchmark",
         "title": "Meson: Run Benchmarks"
+      },
+      {
+        "command": "mesonbuild.restartLanguageServer",
+        "title": "Meson: Restart Language Server"
       }
     ],
     "configuration": {
@@ -187,6 +191,35 @@
             ]
           },
           "description": "Give an object that's merged into the debug configuration provider"
+        },
+        "mesonbuild.languageServer": {
+          "type": "string",
+          "default": "Swift-MesonLSP",
+          "enum": [
+            "Swift-MesonLSP",
+            null
+          ],
+          "description": "Select which language server to use"
+        },
+        "mesonbuild.languageServerPath": {
+          "type": "string",
+          "description": "Binary name or path to language server",
+          "default": ""
+        },
+        "mesonbuild.downloadLanguageServer": {
+          "type": "string",
+          "default": "ask",
+          "enum": [
+            true,
+            false,
+            "ask"
+          ],
+          "enumDescriptions": [
+            "Download the language server",
+            "Do not download the language server",
+            "Ask every time to download the language server"
+          ],
+          "description": "Have VSCode download the language server automatically (MacOS/Windows only for Swift-MesonLSP)"
         }
       }
     },
@@ -338,14 +371,20 @@
     "watch": "tsc -watch -p ./"
   },
   "devDependencies": {
+    "@types/adm-zip": "^0.5.1",
     "@types/node": "^16.11.7",
     "@types/vscode": "^1.1.59",
+    "@types/which": "^3.0.0",
     "husky": "^8.0.3",
     "lint-staged": "^14.0.1",
     "prettier": "^3.0.3",
     "typescript": "^4.4.4"
   },
-  "dependencies": {},
+  "dependencies": {
+    "adm-zip": "^0.5.10",
+    "vscode-languageclient": "^9.0.1",
+    "which": "^4.0.0"
+  },
   "prettier": {
     "proseWrap": "always"
   }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,12 +10,14 @@ import {
   useCompileCommands,
   clearCache,
   checkMesonIsConfigured,
+  getOutputChannel,
 } from "./utils";
 import { DebugConfigurationProvider } from "./configprovider";
 import { testDebugHandler, testRunHandler, rebuildTests } from "./tests";
 import { activateLinters } from "./linters";
 import { activateFormatters } from "./formatters";
-import { TaskQuickPickItem } from "./types";
+import { SettingsKey, TaskQuickPickItem } from "./types";
+import { createLanguageServerClient } from "./lsp/common";
 
 export let extensionPath: string;
 let explorer: MesonProjectExplorer;
@@ -33,9 +35,6 @@ export async function activate(ctx: vscode.ExtensionContext) {
 
   const root = vscode.workspace.workspaceFolders[0].uri.fsPath;
   const buildDir = workspaceRelative(extensionConfiguration("buildFolder"));
-
-  activateLinters(root, ctx);
-  activateFormatters(ctx);
 
   explorer = new MesonProjectExplorer(ctx, root, buildDir);
 
@@ -176,8 +175,7 @@ export async function activate(ctx: vscode.ExtensionContext) {
   );
 
   if (!checkMesonIsConfigured(buildDir)) {
-    const configureOnOpenKey = "configureOnOpen";
-    let configureOnOpen = extensionConfiguration(configureOnOpenKey);
+    let configureOnOpen = extensionConfiguration(SettingsKey.configureOnOpen);
     if (configureOnOpen === "ask") {
       enum Options {
         yes = "Yes",
@@ -196,7 +194,7 @@ export async function activate(ctx: vscode.ExtensionContext) {
           break;
 
         case Options.never:
-          extensionConfigurationSet(configureOnOpenKey, false, vscode.ConfigurationTarget.Workspace);
+          extensionConfigurationSet(SettingsKey.configureOnOpen, false, vscode.ConfigurationTarget.Workspace);
           break;
 
         case Options.yes:
@@ -204,16 +202,71 @@ export async function activate(ctx: vscode.ExtensionContext) {
           break;
 
         case Options.always:
-          extensionConfigurationSet(configureOnOpenKey, true, vscode.ConfigurationTarget.Workspace);
+          extensionConfigurationSet(SettingsKey.configureOnOpen, true, vscode.ConfigurationTarget.Workspace);
           configureOnOpen = true;
           break;
       }
     }
-
-    if (configureOnOpen === true) {
-      runFirstTask("reconfigure");
-    }
   }
+
+  const downloadLanguageServer = extensionConfiguration(SettingsKey.downloadLanguageServer);
+  const server = extensionConfiguration(SettingsKey.languageServer);
+  const shouldDownload = async (downloadLanguageServer: boolean | "ask"): Promise<boolean> => {
+    if (typeof downloadLanguageServer === "boolean") return downloadLanguageServer;
+
+    enum Options {
+      yes = "Yes",
+      no = "Not this time",
+      never = "Never",
+    }
+
+    const response = await vscode.window.showInformationMessage(
+      "Should the extension try to download the language server?",
+      ...Object.values(Options),
+    );
+
+    switch (response) {
+      case Options.yes:
+        extensionConfigurationSet(SettingsKey.downloadLanguageServer, true, vscode.ConfigurationTarget.Global);
+        return true;
+
+      case Options.never:
+        extensionConfigurationSet(SettingsKey.downloadLanguageServer, false, vscode.ConfigurationTarget.Global);
+        return false;
+
+      case Options.no:
+        extensionConfigurationSet(SettingsKey.downloadLanguageServer, "ask", vscode.ConfigurationTarget.Global);
+        return false;
+    }
+
+    return false;
+  };
+
+  let client = await createLanguageServerClient(server, await shouldDownload(downloadLanguageServer), ctx);
+  if (client !== null && server == "Swift-MesonLSP") {
+    ctx.subscriptions.push(client);
+    client.start();
+
+    getOutputChannel().appendLine("Not enabling the muon linter/formatter because Swift-MesonLSP is active.");
+  } else {
+    activateLinters(root, ctx);
+    activateFormatters(ctx);
+  }
+
+  ctx.subscriptions.push(
+    vscode.commands.registerCommand("mesonbuild.restartLanguageServer", async () => {
+      if (client === null) {
+        client = await createLanguageServerClient(server, await shouldDownload(downloadLanguageServer), ctx);
+        if (client !== null) {
+          ctx.subscriptions.push(client);
+          client.start();
+          // TODO: The output line from above about not enabling muon would be good to have here.
+        }
+      } else {
+        client.restart();
+      }
+    }),
+  );
 
   async function pickTask(mode: string) {
     const picker = vscode.window.createQuickPick<TaskQuickPickItem>();
--- a/src/lsp/common.ts
+++ b/src/lsp/common.ts
@@ -1,0 +1,58 @@
+import * as vscode from "vscode";
+import { LanguageServerClient } from ".";
+import { LanguageServer } from "../types";
+import { SwiftMesonLspLanguageClient } from "./swift-mesonlsp";
+import { Uri } from "vscode";
+
+export async function createLanguageServerClient(
+  server: LanguageServer,
+  download: boolean,
+  context: vscode.ExtensionContext,
+): Promise<LanguageServerClient | null> {
+  const serverToClass = (server: LanguageServer) => {
+    switch (server) {
+      case "Swift-MesonLSP":
+        return SwiftMesonLspLanguageClient;
+      default:
+        return null;
+    }
+  };
+
+  const klass = serverToClass(server);
+  if (klass == null) {
+    return null;
+  }
+  if (!klass.supportsSystem()) {
+    vscode.window.showErrorMessage("The configured language server does not support the current system.");
+    return null;
+  }
+
+  let languageServerPath = LanguageServerClient.resolveLanguageServerPath(server, context);
+  if (languageServerPath === null) {
+    if (klass.artifact() == null) {
+      enum Options {
+        open = "Open documentation in browser",
+      }
+      const response = await vscode.window.showErrorMessage(
+        "This language server supports your systen, but provides no artifacts for automatic setup",
+        ...Object.values(Options),
+      );
+      if (response == Options.open) {
+        vscode.env.openExternal(Uri.parse(klass.setupURL));
+      }
+      return null;
+    }
+    if (download) {
+      languageServerPath = await klass.download(server, klass.version, context);
+      if (languageServerPath === null) {
+        vscode.window.showErrorMessage("Failed to download the language server.");
+        return null;
+      }
+    } else {
+      vscode.window.showErrorMessage("Failed to find a language server on the system.");
+      return null;
+    }
+  }
+
+  return new klass(languageServerPath, context);
+}
--- a/src/lsp/index.ts
+++ b/src/lsp/index.ts
@@ -1,0 +1,196 @@
+import * as Admzip from "adm-zip";
+import * as which from "which";
+import * as https from "https";
+import * as crypto from "crypto";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import * as vscode from "vscode";
+import {
+  Executable,
+  LanguageClient,
+  LanguageClientOptions,
+  ServerOptions,
+  TransportKind,
+} from "vscode-languageclient/node";
+import * as storage from "../storage";
+import { LanguageServer } from "../types";
+
+export abstract class LanguageServerClient {
+  private static readonly clientOptions: LanguageClientOptions = {
+    documentSelector: ["meson", { scheme: "file", language: "meson" }],
+  };
+  private ls: LanguageClient | null = null;
+  private readonly context: vscode.ExtensionContext;
+
+  protected languageServerPath: vscode.Uri | null;
+  readonly server: LanguageServer;
+
+  static readonly repoURL: string;
+  static readonly setupURL: string;
+  static readonly version: string;
+
+  protected abstract get debugExe(): Executable;
+  protected abstract get runExe(): Executable;
+
+  protected constructor(server: LanguageServer, languageServerPath: vscode.Uri, context: vscode.ExtensionContext) {
+    this.server = server;
+    this.languageServerPath = languageServerPath;
+    this.context = context;
+  }
+
+  private static cachedLanguageServer(server: LanguageServer, context: vscode.ExtensionContext): vscode.Uri | null {
+    const uri = vscode.Uri.joinPath(
+      storage.uri(storage.Location.LSP, context),
+      `${server}${os.platform() === "win32" ? ".exe" : ""}`,
+    );
+
+    return fs.existsSync(uri.fsPath) ? uri : null;
+  }
+
+  private static async computeFileHash(filePath: string): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
+      const hash = crypto.createHash("sha256");
+      const stream = fs.createReadStream(filePath);
+
+      stream.on("error", reject);
+
+      stream.on("data", (data) => {
+        hash.update(data);
+      });
+
+      stream.on("end", () => {
+        resolve(hash.digest("hex"));
+      });
+    });
+  }
+
+  private static async fetch(url: string, dest: string): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const file = fs.createWriteStream(dest);
+      const request = https.get(url, (response) => {
+        if (response.statusCode === 302 || response.statusCode === 301 || response.statusCode === 302) {
+          LanguageServerClient.fetch(response.headers.location!, dest).then(resolve).catch(reject);
+        } else {
+          response.pipe(file);
+
+          file.on("finish", () => {
+            file.close((_) => {
+              resolve();
+            });
+          });
+
+          file.on("error", (err) => {
+            vscode.window.showErrorMessage(`Error writing to file: ${err}`);
+            reject(err);
+          });
+        }
+      });
+
+      request.on("error", (err) => {
+        vscode.window.showErrorMessage(`Request error: ${err}`);
+        reject(err);
+      });
+    });
+  }
+
+  static async download(
+    server: LanguageServer,
+    version: string,
+    context: vscode.ExtensionContext,
+  ): Promise<vscode.Uri | null> {
+    const lspDir = storage.uri(storage.Location.LSP, context).fsPath;
+    const artifact = this.artifact();
+    if (artifact === null) return null;
+
+    fs.rmSync(lspDir, { recursive: true, force: true });
+    fs.mkdirSync(lspDir, { recursive: true });
+
+    let uri: vscode.Uri | null = null;
+    const tmpPath = path.join(os.tmpdir(), `vscode-meson-${server}-${Date.now()}.zip`);
+
+    try {
+      await LanguageServerClient.fetch(artifact.url, tmpPath);
+      let hash = await this.computeFileHash(tmpPath);
+      if (hash !== artifact.hash) {
+        vscode.window.showErrorMessage(`Invalid hash: Expected ${artifact.hash}, got ${hash}.`);
+        fs.unlinkSync(tmpPath);
+        return null;
+      }
+
+      const zip = new Admzip(tmpPath);
+      zip.extractAllTo(lspDir);
+      const binary = path.join(lspDir, server!);
+      if (os.platform() != "win32") fs.chmodSync(binary, 0o755);
+      const versionFile = path.join(lspDir, "version");
+      fs.writeFileSync(versionFile, version);
+
+      vscode.window.showInformationMessage("Language server was downloaded.");
+      uri = vscode.Uri.from({ scheme: "file", path: binary });
+      fs.unlinkSync(tmpPath);
+      return uri;
+    } catch (err) {
+      vscode.window.showErrorMessage(JSON.stringify(err));
+      fs.unlinkSync(tmpPath);
+    }
+    return uri;
+  }
+
+  static resolveLanguageServerPath(server: LanguageServer, context: vscode.ExtensionContext): vscode.Uri | null {
+    const config = vscode.workspace.getConfiguration("mesonbuild");
+    if (config["languageServerPath"] !== null && config["languageServerPath"] != "")
+      return vscode.Uri.from({ scheme: "file", path: config["languageServerPath"] });
+
+    const cached = LanguageServerClient.cachedLanguageServer(server, context);
+    if (cached !== null) return cached;
+
+    const binary = which.sync(server!, { nothrow: true });
+    if (binary !== null) return vscode.Uri.from({ scheme: "file", path: binary });
+
+    return null;
+  }
+
+  protected static supportsSystem(): boolean {
+    return true;
+  }
+
+  protected static artifact(): { url: string; hash: string } | null {
+    return null;
+  }
+
+  dispose() {
+    if (this.ls !== null) {
+      this.ls.stop();
+      this.ls = null;
+    }
+  }
+
+  restart(): void {
+    this.dispose();
+    this.languageServerPath = LanguageServerClient.resolveLanguageServerPath(this.server, this.context);
+    if (this.languageServerPath === null) {
+      vscode.window.showErrorMessage(
+        "Failed to restart the language server because a binary was not found and could not be downloaded",
+      );
+    } else {
+      this.start();
+    }
+  }
+
+  start(): void {
+    const serverOptions: ServerOptions = {
+      run: this.runExe,
+      debug: this.debugExe,
+      transport: TransportKind.stdio,
+    };
+
+    this.ls = new LanguageClient(
+      this.server!,
+      `Meson Language Server (${this.server})`,
+      serverOptions,
+      LanguageServerClient.clientOptions,
+      true,
+    );
+    this.ls.start();
+  }
+}
--- a/src/lsp/swift-mesonlsp.ts
+++ b/src/lsp/swift-mesonlsp.ts
@@ -1,0 +1,71 @@
+import * as os from "os";
+import * as vscode from "vscode";
+
+import { LanguageServerClient } from "../lsp";
+import { Executable } from "vscode-languageclient/node";
+
+export class SwiftMesonLspLanguageClient extends LanguageServerClient {
+  private static artifacts: { [key: string]: { name: string; hash: string } } = {
+    "win32-x64": {
+      name: "Swift-MesonLSP-win64.zip",
+      hash: "81e4c433623b344d8e321a0e48426b3f83844b108d099a8cac6993c546084b66",
+    },
+    "darwin-x64": {
+      name: "Swift-MesonLSP-macos12.zip",
+      hash: "9073875ba3eb75fe0c2123e4d0e09c8bccde271c907e74fc84110c85662e4a49",
+    },
+    "linux-x64": {
+      name: "Swift-MesonLSP.zip",
+      hash: "2b9359e9cff8cec40279f78c62ef8eb94693e26e21716f82293cf96ad1f723aa",
+    },
+  };
+
+  static override repoURL: string = "https://github.com/JCWasmx86/Swift-MesonLSP";
+  static override setupURL: string = "https://github.com/JCWasmx86/Swift-MesonLSP/tree/main/Docs";
+  static override version: string = "2.4.4";
+
+  get runExe(): Executable {
+    return {
+      command: this.languageServerPath!.fsPath,
+      args: ["--lsp"],
+    };
+  }
+
+  get debugExe(): Executable {
+    return {
+      command: this.languageServerPath!.fsPath,
+      args: ["--lsp"],
+    };
+  }
+
+  constructor(languageServerPath: vscode.Uri, context: vscode.ExtensionContext) {
+    super("Swift-MesonLSP", languageServerPath, context);
+  }
+
+  static override artifact(): { url: string; hash: string } | null {
+    const arch = os.arch();
+    const platform = os.platform();
+    if (arch !== "x64") return null;
+
+    const artifact = SwiftMesonLspLanguageClient.artifacts[`${platform}-${arch}`];
+    return {
+      url: `${SwiftMesonLspLanguageClient.repoURL}/releases/download/v${SwiftMesonLspLanguageClient.version}/${artifact.name}`,
+      hash: artifact.hash,
+    };
+  }
+
+  static override supportsSystem(): boolean {
+    const arch = os.arch();
+    if (arch != "x64") return false;
+
+    const platform = os.platform();
+    switch (platform) {
+      case "darwin":
+      case "linux":
+      case "win32":
+        return true;
+      default:
+        return false;
+    }
+  }
+}
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -1,0 +1,9 @@
+import * as vscode from "vscode";
+
+export enum Location {
+  LSP = "lsp",
+}
+
+export function uri(location: Location, context: vscode.ExtensionContext): vscode.Uri {
+  return vscode.Uri.joinPath(context.globalStorageUri, location);
+}
--- a/src/types.ts
+++ b/src/types.ts
@@ -8,6 +8,8 @@ export type LinterConfiguration = {
   enabled: boolean;
 };
 
+export type LanguageServer = "Swift-MesonLSP" | null;
+
 export interface ExtensionConfiguration {
   configureOnOpen: boolean | "ask";
   configureOptions: string[];
@@ -27,6 +29,9 @@ export interface ExtensionConfiguration {
     muonConfig: string | null;
   };
   debugOptions: object;
+  languageServer: LanguageServer;
+  languageServerPath: string;
+  downloadLanguageServer: boolean | "ask";
 }
 
 export interface TaskQuickPickItem extends vscode.QuickPickItem {
@@ -119,3 +124,9 @@ export type Targets = Target[];
 export type BuildOptions = BuildOption<any>[];
 export type Dependencies = Dependency[];
 export type Tests = Test[];
+
+export enum SettingsKey {
+  downloadLanguageServer = "downloadLanguageServer",
+  languageServer = "languageServer",
+  configureOnOpen = "configureOnOpen",
+}
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,55 +2,84 @@
 # yarn lockfile v1
 
 
-"@types/node@^16.11.7":
+"@types/adm-zip@^0.5.1":
+  version "0.5.2"
+  resolved "https://registry.npmjs.org/@types/adm-zip/-/adm-zip-0.5.2.tgz"
+  integrity sha512-33OTTnnW3onOE6HJuoqsi7T7Ojupz7zO/Vs5ddRNVCYQnu4lg05RqH/pr9eidHGvGyYfdO4uPO9cvegAMixBCQ==
+  dependencies:
+    "@types/node" "*"
+
+"@types/node@*", "@types/node@^16.11.7":
   version "16.11.7"
   resolved "https://registry.npmjs.org/@types/node/-/node-16.11.7.tgz"
   integrity sha512-QB5D2sqfSjCmTuWcBWyJ+/44bcjO7VbjSbOE0ucoVbAsSNQc4Lt6QkgkVXkTDwkL4z/beecZNDvVX15D4P8Jbw==
 
 "@types/vscode@^1.1.59":
-  version "1.64.0"
-  resolved "https://registry.npmjs.org/@types/vscode/-/vscode-1.64.0.tgz"
-  integrity sha512-bSlAWz5WtcSL3cO9tAT/KpEH9rv5OBnm93OIIFwdCshaAiqr2bp1AUyEwW9MWeCvZBHEXc3V0fTYVdVyzDNwHA==
+  version "1.81.0"
+  resolved "https://registry.npmjs.org/@types/vscode/-/vscode-1.81.0.tgz"
+  integrity sha512-YIaCwpT+O2E7WOMq0eCgBEABE++SX3Yl/O02GoMIF2DO3qAtvw7m6BXFYsxnc6XyzwZgh6/s/UG78LSSombl2w==
+
+"@types/which@^3.0.0":
+  version "3.0.0"
+  resolved "https://registry.npmjs.org/@types/which/-/which-3.0.0.tgz"
+  integrity sha512-ASCxdbsrwNfSMXALlC3Decif9rwDMu+80KGp5zI2RLRotfMsTv7fHL8W8VDp24wymzDyIFudhUeSCugrgRFfHQ==
+
+adm-zip@^0.5.10:
+  version "0.5.10"
+  resolved "https://registry.npmjs.org/adm-zip/-/adm-zip-0.5.10.tgz"
+  integrity sha512-x0HvcHqVJNTPk/Bw8JbLWlWoo6Wwnsug0fnYYro1HBrjxZ3G7/AZk7Ahv8JwDe1uIcz8eBqvu86FuF1POiG7vQ==
 
 ansi-escapes@^5.0.0:
   version "5.0.0"
-  resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-5.0.0.tgz#b6a0caf0eef0c41af190e9a749e0c00ec04bb2a6"
+  resolved "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-5.0.0.tgz"
   integrity sha512-5GFMVX8HqE/TB+FuBJGuO5XG0WrsA6ptUqoODaT/n9mmUaZFkqnBueB4leqGBCmrUHnCnC4PCZTCd0E7QQ83bA==
   dependencies:
     type-fest "^1.0.2"
 
 ansi-regex@^6.0.1:
   version "6.0.1"
-  resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-6.0.1.tgz#3183e38fae9a65d7cb5e53945cd5897d0260a06a"
+  resolved "https://registry.npmjs.org/ansi-regex/-/ansi-regex-6.0.1.tgz"
   integrity sha512-n5M855fKb2SsfMIiFFoVrABHJC8QtHwVx+mHWP3QcEqBHYienj5dHSgjbxtC0WEZXYt4wcD6zrQElDPhFuZgfA==
 
 ansi-styles@^6.0.0, ansi-styles@^6.1.0:
   version "6.2.1"
-  resolved "https://registry.yarnpkg.com/ansi-styles/-/ansi-styles-6.2.1.tgz#0e62320cf99c21afff3b3012192546aacbfb05c5"
+  resolved "https://registry.npmjs.org/ansi-styles/-/ansi-styles-6.2.1.tgz"
   integrity sha512-bN798gFfQX+viw3R7yrGWRqnrN2oRkEkUjjl4JNn4E8GxxbjtG3FbrEIIY3l8/hrwUwIeCZvi4QuOTP4MErVug==
+
+balanced-match@^1.0.0:
+  version "1.0.2"
+  resolved "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.2.tgz"
+  integrity sha512-3oSeUO0TMV67hN1AmbXsK4yaqU7tjiHlbxRDZOpH0KW9+CeX4bRAaX0Anxt0tx2MrpRpWwQaPwIlISEJhYU5Pw==
+
+brace-expansion@^2.0.1:
+  version "2.0.1"
+  resolved "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.0.1.tgz"
+  integrity sha512-XnAIvQ8eM+kC6aULx6wuQiwVsnzsi9d3WxzV3FpWTGA19F621kwdbsAcFKXgKUHZWsy+mY6iL1sHTxWEFCytDA==
+  dependencies:
+    balanced-match "^1.0.0"
 
 braces@^3.0.2:
   version "3.0.2"
-  resolved "https://registry.yarnpkg.com/braces/-/braces-3.0.2.tgz#3454e1a462ee8d599e236df336cd9ea4f8afe107"
+  resolved "https://registry.npmjs.org/braces/-/braces-3.0.2.tgz"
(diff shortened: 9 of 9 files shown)


JCWasmx86:
I think this is at least partially ready for review (Sadly I wasn't able to find a way - even with help of others - to download a file that worked for me. It would always end up empty)

Automatic updates are currently not implemented as I don't know what the best way would be.

This adds a few dependencies:
- [which](https://www.npmjs.com/package/which) to find the language server on the path
- [adm-zip](https://www.npmjs.com/package/adm-zip) to extract the downloaded zip

lsp.ts is based a bit on the [Vala VSCode extension (MIT)](https://github.com/vala-lang/vala-vscode)

package.json
```diff
@@ -177,6 +177,16 @@
             ]
           },
           "description": "Give an object that's merged into the debug configuration provider"
+        },
+        "mesonbuild.languageServerPath": {
```
tristan957, line 181: It is possible that there could be multiple language server implementations in the future. Perhaps these settings could take inspiration from the Python extension because I think it supports multiple?

package.json
```diff
@@ -177,6 +177,16 @@
...
+          "type": "string",
+          "description": "Binary name or path to language server",
+          "default": ""
+        },
+        "mesonbuild.neverDownloadLanguageServer": {
```
tristan957, line 186: I would say drop the never here

package.json
```diff
@@ -323,7 +333,13 @@
   "devDependencies": {
     "@types/node": "^16.11.7",
     "@types/vscode": "^1.1.59",
-    "typescript": "^4.4.4"
+    "typescript": "^4.4.4",
```
tristan957, line 336: nit pick, but maybe alphabetize.

src/lsp.ts
```diff
@@ -0,0 +1,77 @@
...
+} from "./utils";
+
+import * as which from "which"
+
+export class MesonLanguageClient {
+  config: WorkspaceConfiguration
+  ls: LanguageClient | null = null
+
+  constructor(_context: ExtensionContext) {
+    this.config = workspace.getConfiguration('vala');
```
tristan957, line 26: Is 'vala' right here?

src/lsp.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.startLanguageServer(serverModule);
+  }
+
+  get languageServerPath(): string | null {
+    return findLanguageServer() || this.config.languageServerPath || which.sync("Swift-MesonLSP", { nothrow: true })
+  }
+
+  dispose() {
+    if (this.ls) {
+      this.ls!.stop()
```
tristan957, line 72: Is the exclamation mark really necessary here?

tristan957:
The clangd, codeql, rust analyzer extensions might be good to look at for inspiration on auto updating.

package.json
JCWasmx86: That's implicitly supported. You just have to point it to an executable that `--lsp --stdio` as parameters. Sure not optimal

tristan957: I think this extension having a blessed default (yours) makes the most sense. Then someone could change the setting if another language server ever cropped up.

I don't think any pop up is necessary other than asking do you want to download the latest version of the configured language server.

src/lsp.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.config = workspace.getConfiguration("mesonbuild");
+    let serverModule = this.languageServerPath;
+
+    if (serverModule == null)
+      return;
+    this.startLanguageServer(serverModule);
+  }
+
+  startLanguageServer(serverModule: string): void {
+    let clientOptions: LanguageClientOptions = {
```
tristan957, line 35: Seems like there could be more "const" instead of "let" here, but I am definitely no expert :).

src/utils.ts
```diff
@@ -149,6 +154,135 @@ export function extensionConfigurationSet<
   return getConfiguration().update(key, value, target);
 }
 
+export function canDownloadLanguageServer(): boolean {
+  const platform = os.platform()
+  if (platform != "win32" && platform != "darwin") {
+    return false
```
tristan957, line 160: Style: Inconsistency when if statements are only one line.

src/utils.ts
```diff
@@ -149,6 +154,135 @@ export function extensionConfigurationSet<
...
+}
+
+function createDownloadURL(): string {
+  const platform = os.platform()
+  const filename = platform == "win32" ? "Swift-MesonLSP-win64.zip" : "Swift-MesonLSP-macos12.zip"
+  return `https://github.com/JCWasmx86/Swift-MesonLSP/releases/download/v2.0/${filename}`
+}
+
+function createHashForLanguageServer(): string {
+  const platform = os.platform()
```
tristan957, line 173: A switch would probably be better here and other places, but could be wrong.

src/utils.ts
```diff
@@ -149,6 +154,135 @@ export function extensionConfigurationSet<
...
+      }).catch((err: Error) => vscode.window.showErrorMessage(err.message));
+    }).catch((err: Error) => vscode.window.showErrorMessage(err.message));
+
+  } catch (err) {
+    vscode.window.showErrorMessage(JSON.stringify(err));
+    return;
+  }
+}
+
+
```
tristan957, line 238: extra new line

src/utils.ts
```diff
@@ -149,6 +154,135 @@ export function extensionConfigurationSet<
...
+      force: true
+    }
+  )
+  fs.mkdirSync(
+    lspDir,
+    { recursive: true }
+  )
+  const tmpPath = path.join(os.tmpdir(), `lsp-${Date.now()}.zip`);
+  try {
+    const promise = downloadFile(createDownloadURL(), tmpPath);
```
tristan957, line 212: Any reason not to just use await here?

src/utils.ts
```diff
@@ -149,6 +154,135 @@ export function extensionConfigurationSet<
...
+    });
+    request.on('error', (err) => {
+      vscode.window.showErrorMessage(`Request error: ${err}`);
+      reject(err);
+    });
+  });
+}
+
+const getExtensionDir = (): string => {
+  return path.join(os.homedir(), ".vscode-meson");
```
tristan957, line 268: Doesn't VSCode give extensions a place to store data?

src/utils.ts
```diff
@@ -149,6 +154,135 @@ export function extensionConfigurationSet<
...
+  const platform = os.platform()
+  if (platform != "win32" && platform != "darwin") {
+    return false
+  }
+  const arch = os.arch()
+  return arch == "x64" && findLanguageServer() == null;
+}
+
+function createDownloadURL(): string {
+  const platform = os.platform()
```
tristan957, line 167: I think some of these functions should be getters that are only computed the first time. Perhaps creating a `MesonLanguageSever` abstract class, and then you create an implementation for your client. Your client would obviously be the only implementation.

```typescript
abstract class MesonLanguageServer {
  name: string;
  serverOptions: ServerOptions;

  constructor() {
    // This would be your current setupLanguageServer maybe?
  }

  get downloadInfo(): [url: string, hash: string] | null {
    return null;
  }

  private dispose(): void {

  }

  restart(): void {
    // ...
  }
}
```

Hopefully you get what I am saying

package.json
```diff
@@ -177,6 +177,21 @@
...
+        },
+        "mesonbuild.downloadLanguageServer": {
+          "type": "string",
+          "default": "ask",
+          "enum": [
+            "ask",
+            "never",
+            "yes"
+          ],
+          "description": "Have VSCode download the language server automatically (MacOS/Windows only)"
```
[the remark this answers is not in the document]
JCWasmx86: I'm no expert, but you have to consider e.g. NixOS with its weird fixed up paths, different glibc versions etc. (All the reasons flatpak exists)

That's why my/dcbaker's idea (No sure from whom this idea was😅) was to just use distro packages. For Arch I setup an AUR, for Fedora a COPR. dcbaker wanted to write a nix expression. For Ubuntu/Debian I wanted to setup an APT repo sometime.

src/lsp.ts
```diff
@@ -0,0 +1,77 @@
...
+      command: serverModule,
+      args: ["--lsp"],
+    };
+    let serverOptions: ServerOptions = {
+      run: runExe,
+      debug: debugExe,
+      transport: TransportKind.stdio
+    };
+
+    this.ls = new LanguageClient("Swift-MesonLSP", "Meson Language Server", serverOptions, clientOptions, true);
```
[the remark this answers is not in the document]
JCWasmx86: Tbh I like the name, it says everything "language server for meson" and "written in Swift" or "It's fast", depending if you know Swift ^^

src/utils.ts
JCWasmx86: I'm sure there is one, but I didn't find it :)

dcbaker:
I'm sorry I've been really behind on this, lots of RL stuff right now. I will try to get a nix expression written so we can test the distro provided mechanism

JCWasmx86:
No worries. I've already implemented that it searches in this order:

1. The extension internal directory (Where the language server was downloaded)
2. The executable from the config
3. path

package.json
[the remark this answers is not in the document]
JCWasmx86: Did search a bit more: Linking with musl is currently (22 Jun. 2022) not possible. So using the package managers is the best action

src/utils.ts
tristan957: https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage

JCWasmx86: Yes, I found it already ^^ Nevertheless thanks

JCWasmx86:
For updates I found two ways:

### Using Github releases
Just download the latest artifacts from Github.
### Maintain separate JSON
E.g. have some JSON file
```
{
  "versions": {
    "2.0": {
      "win32-x64": {
        "hash": "1122334455",
        "url": "https://someurl"
      },
    }
  }
}
```
The keys could be a concatenation of os.platform() and os.arch(), thus dynamically allowing other os/arch combinations.

package.json
JCWasmx86: Had an idea: Maybe instead of downloading, setup everything automatically, if a supported distribution is found. That would require root rights, but I'm sure this could work in some way

dcbaker:
There's some rebase issues that will need to be sorted, but the code itself looks good to me. I'm going to write a nix expression and give this a test drive. I forgot to look, did you add an entry for this to the `CHANGELOG`?

JCWasmx86:
I can resolve the conflicts later. I've not added a changelog section yet.

JCWasmx86:
I've resolved the conflicts and added an section to the changelog including an optional very concise help for Linux users

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## next
 
+- Add integration for [Swift-MesonLSP](https://github.com/JCWasmx86/Swift-MesonLSP).
+  vscode-meson will automatically download it on Windows and MacOS. Linux users should
+  follow the [docs](https://github.com/JCWasmx86/Swift-MesonLSP/tree/main/Docs)
```
tristan957, line 7: Add a newline.

src/lsp.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  get languageServerPath(): string | null {
+    return findLanguageServer(this) || this.config.languageServerPath || which.sync(this.executableName, { nothrow: true })
+  }
+
+  async setupLanguageServer() {
+    setupLanguageServer(this)
+  }
+
+  dispose() {
```
tristan957, line 52: It has been a while since I wrote TS. Do you have to mark this private?

src/lsp.ts
```diff
@@ -0,0 +1,59 @@
...
+      return;
+    this.startLanguageServer(serverModule);
+  }
+
+  canDownloadLanguageServer(): boolean {
+    return canDownloadLanguageServer(this)
+  }
+
+  get languageServerPath(): string | null {
+    return findLanguageServer(this) || this.config.languageServerPath || which.sync(this.executableName, { nothrow: true })
```
tristan957, line 45: The `this.config.languageServerPath` should probably come first right?

src/utils.ts
```diff
@@ -123,6 +128,118 @@ export function extensionConfigurationSet<
   return getConfiguration().update(key, value, target);
 }
 
+export function canDownloadLanguageServer(client: LanguageServerClient): boolean {
```
tristan957, line 131: Can't you use LanguageServerClient::donwloadInfo() instead of having this function? Perhaps it just got left by accident

src/lsp.ts
JCWasmx86: Not sure, but making it private breaks nothing

tristan957: I only say that because it is only called in finalize.

tristan957:
Now the CI can run since I have the relevant permissions :)

src/utils.ts
```diff
@@ -123,6 +128,110 @@ export function extensionConfigurationSet<
   return getConfiguration().update(key, value, target);
 }
 
+export function findLanguageServer(client: LanguageServerClient): string | null {
+  const platform = os.platform();
+  const executableName = client.executableName
+  if (platform != "win32" && platform != "darwin")
```
tristan957, line 134: Some of this logic seems specific to your language server. Might try to find a way to move this into the abstract class somehow?

src/lsp.ts
```diff
@@ -0,0 +1,58 @@
...
+  canDownloadLanguageServer(): boolean {
+    return this.downloadInfo != null && findLanguageServer(this) == null
+  }
+
+  get languageServerPath(): string | null {
+    return this.config.languageServerPath || findLanguageServer(this) || which.sync(this.executableName, { nothrow: true })
+  }
+
+  async setupLanguageServer() {
+    setupLanguageServer(this)
```
tristan957, line 48: Seems like you should just inline the other function to me. No need to call an out of band function.

src/lsp.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  canDownloadLanguageServer(): boolean {
+    return this.downloadInfo != null && findLanguageServer(this) == null
+  }
+
+  get languageServerPath(): string | null {
+    return this.config.languageServerPath || findLanguageServer(this) || which.sync(this.executableName, { nothrow: true })
+  }
+
+  async setupLanguageServer() {
```
tristan957, line 47: I would mark the return type as `Promise<void>`

src/utils.ts
JCWasmx86: Maybe some computed properties that returns all supported OS/Arch combinations?

tristan957: Perhaps. I am thinking of a `LanguageServerClient::find()` function or some sort. You could mark the function as not abstract, and then use the computed property like you said. Might not need each subclass to write their own impl that way.

JCWasmx86:
utils.ts is no unchanged again compared to main, but broke something somewhere as mkdirSync fails now with ENOENT

JCWasmx86:
Should work again. Mixed up path of an URL and the fsPath

src/lsp.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+      });
+      request.on('error', (err) => {
+        vscode.window.showErrorMessage(`Request error: ${err}`);
+        reject(err);
+      });
+    });
+  }
+
+  findLanguageServer(): string | null {
```
tristan957, line 134: I think this can be made private?

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,64 @@
...
+            "c10999081cef8b56ea164d9a170470c268045254829e48bf4ac137a07a8a874c"]
+  }
+
+  supportsSystem(os: string, arch: string): boolean {
+    if (arch != "x64")
+      return false;
+    return os == "win32" || os == "darwin";
+  }
+
+  startLanguageServer(serverModule: string): void {
```
tristan957, line 42: I think this function can be a private function to the abstract parent class. The `clientOptions` should probably be a private member of that base class. I don't see it changing for other potential clients. Do you?

I think `runExe` and `debugExe` could probably be abstract getters on the base class? Then `serverOptions` could be made using the abstract getters.

Overall, I think this design is coming together quite nicely.

src/lsp.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+    )
+    fs.mkdirSync(
+      lspDir,
+      { recursive: true }
+    )
+    const tmpPath = path.join(os.tmpdir(), `lsp-${Date.now()}.zip`);
+    try {
+      this.downloadFile(downloadInfo[0], tmpPath).then(() => {
+        const hashPromise = this.computeFileHash(tmpPath);
```
tristan957, line 69: what do you think about moving the hash check to the download function?

src/lsp.ts
```diff
@@ -0,0 +1,160 @@
...
+        force: true
+      }
+    )
+    fs.mkdirSync(
+      lspDir,
+      { recursive: true }
+    )
+    const tmpPath = path.join(os.tmpdir(), `lsp-${Date.now()}.zip`);
+    try {
+      this.downloadFile(downloadInfo[0], tmpPath).then(() => {
```
tristan957, line 68: Should we clean up the file if anything of the following code `reject`s

src/lsp.ts
```diff
@@ -0,0 +1,160 @@
...
+    this.context = context
+    this.config = workspace.getConfiguration("mesonbuild");
+    const serverModule = this.languageServerPath;
+
+    if (serverModule == null)
+      return;
+    this.startLanguageServer(serverModule);
+  }
+
+  abstract startLanguageServer(serverModule: string): void
```
tristan957, line 29: Trailing semicolons!

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,64 @@
...
+  Executable,
+  TransportKind
+} from "vscode-languageclient/node";
+import {
+  ExtensionContext,
+} from "vscode"
+import { LanguageServerClient } from "../lsp";
+
+export class SwiftMesonLspLanguageClient extends LanguageServerClient {
+  ls: LanguageClient | null = null
```
tristan957, line 16: I don't think is necessary? Looks like you already have one in the parent class.

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (arch != "x64")
+      return null;
+    if (platform == "win32")
+      return ["https://github.com/JCWasmx86/Swift-MesonLSP/releases/download/v2.0/Swift-MesonLSP-win64.zip",
+              "fc0d930eb67525309291907c07df8655920cdc651f17f2320d09cc5271a23876"]
+    return ["https://github.com/JCWasmx86/Swift-MesonLSP/releases/download/v2.0/Swift-MesonLSP-macos12.zip",
+            "c10999081cef8b56ea164d9a170470c268045254829e48bf4ac137a07a8a874c"]
+  }
+
+  supportsSystem(os: string, arch: string): boolean {
```
tristan957, line 36: Doesn't your language server support linux?

JCWasmx86: Yes but no  setup here in the extension due to the issues discussed elsewhere

src/lsp/swift-mesonlsp.ts
JCWasmx86: debugExe/runExe depend on `serverModule`, so best I can do is a method that e.g. returns an array of executables. Or we could mandate, that all implementations have to accept `--lsp`?

src/lsp.ts
JCWasmx86: Not a fan of it, as it won't be composable anymore. One function = one functionality

src/lsp.ts
JCWasmx86: Not a fan of it, as it won't be composable anymore. One function = one functionality

src/lsp/swift-mesonlsp.ts
JCWasmx86: I have two ideas regarding this:
- Automatically setup the package manager on supported OS. This requires elevation to root (
- Upload more binaries for different distributions to Github Releases (Compiles on Ubuntu 18.04/20.04/22.04, Fedora, Arch). But it would be outside of the package manager and at least I would be annoyed it some extension downloads some binary albeit Linux has a package manager.

JCWasmx86:
On my TODO list is still:  "Should muon and Swift-MesonLSP be XOR? Swift-MesonLSP uses muon for formatting and duplicates a lot of work from there. So you would have e.g. duplicate diagnostics"

Should I simply disable the muon integration, if a running language server is detected?

src/lsp.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  abstract get repoURL(): string;
+
+  restart(): void {
+    this.dispose();
+    const serverModule = this.languageServerPath;
+
+    if (serverModule == null)
+      return;
+    this.startLanguageServer(serverModule);
```
tristan957, line 44: Why do you need to pass `serverModule` here? The function you are classing has access to the `this.languageServerPath`.

src/lsp.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  private startLanguageServer(serverModule: string) {
+    const runExe: Executable = {
+      command: serverModule,
+      args: ["--lsp"],
```
tristan957, line 50: After you fix the previous comment do the abstract getter for exposing the run and debug exes. The subclasses can access this.languageServerPath and not need to take serverModule as an argument anymore

src/lsp.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (serverModule == null)
+      return;
+    this.startLanguageServer(serverModule);
+  }
+
+  abstract supportsSystem(os: string, arch: string): boolean
+
+  abstract get downloadInfo(): [url: string, hash: string] | null;
+
+  abstract get repoURL(): string;
```
tristan957, line 36: I don't think this has to be a get?

It can be

```
abstract repoURL: string;
```

src/extension.ts
```diff
@@ -189,10 +190,44 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
+  }
 
-    if (configureOnOpen === true) {
-      runFirstTask("reconfigure");
+  const downloadLanguageServerKey = "downloadLanguageServer";
+  const downloadLanguageServer = extensionConfiguration(downloadLanguageServerKey);
+  const languageServerKey = "languageServer"
+  const lsKey = extensionConfiguration(languageServerKey);
+  const possibleServers = {
+    "Swift-MesonLSP": new SwiftMesonLspLanguageClient(ctx)
```
tristan957, line 200: Don't preconstruct the clients. Wasted resources.

```
const client = possibleServers[key](ctx);

**Comment by JCWasmx86** [comment]
Would moving the calls
```
activateLinters(root, ctx);
activateFormatters(ctx);
```
in extension.ts either below the question whether you want to install a language server/if a language server is installed or should I move my code over these two lines?

**Comment by tristan957** [comment]
Just pick one and run with it imo.

File: src/lsp.ts [hunk 33]
```diff
@@ -0,0 +1,180 @@
...
+  config: WorkspaceConfiguration;
+  ls: LanguageClient | null = null;
+  executableName: string;
+  context: ExtensionContext;
+  private clientOptions: LanguageClientOptions = {
+    documentSelector: ["meson", { "scheme": "file", language: "meson" }]
+  };
+  abstract repoURL: string;
+  abstract get debugExe(): Executable
+  abstract get runExe(): Executable
```
tristan957, line 22: Some of these should probably be marked as private/protected. Also ending semicolons

src/lsp.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  constructor(executableName: string, context: ExtensionContext) {
+    this.executableName = executableName;
+    this.context = context;
+    this.config = workspace.getConfiguration("mesonbuild");
+    const serverModule = this.languageServerPath;
```
tristan957, line 28: get rid of this temp variable

src/lsp.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  private startLanguageServer() {
+    const serverOptions: ServerOptions = {
+      run: this.runExe,
+      debug: this.debugExe,
+      transport: TransportKind.stdio
+    };
+
+    this.ls = new LanguageClient("Swift-MesonLSP", "Meson Language Server", serverOptions, this.clientOptions, true);
```
tristan957, line 54: This needs to be made more generic :)

src/lsp.ts
```diff
@@ -0,0 +1,180 @@
...
+    this.ls.start();
+  }
+
+  canDownloadLanguageServer(): boolean {
```
tristan957, line 58: I think the logic in this function is strange. Whether you can download the language server is different that if you have currently found it on the current system

src/lsp.ts
```diff
@@ -0,0 +1,180 @@
...
+    return this.downloadInfo != null && this.findLanguageServer() == null
+  }
+
+  get languageServerPath(): string | null {
+    return this.config.languageServerPath || this.findLanguageServer() || which.sync(this.executableName, { nothrow: true })
```
tristan957, line 63: Is it possible to just resolve the language server path in the constructor?

src/lsp.ts
```diff
@@ -0,0 +1,180 @@
...
+      return fullpath;
+    }
+    return null;
+  }
+
+  private getExtensionLSPDir(ctx: vscode.ExtensionContext): string {
+    return path.join(ctx.globalStorageUri.fsPath, "lsp");
+  }
+
+  public dispose() {
```
tristan957, line 174: this should be made private

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,47 @@
...
+  get runExe(): Executable {
+    return {
+    command: this.languageServerPath,
+    args: ["--lsp"]
+    }
+  }
+  get debugExe(): Executable {
+    return {
+    command: this.languageServerPath,
+    args: ["--lsp"]
```
tristan957, line 20: formatting

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  get downloadInfo(): [url: string, hash: string] | null {
+    const platform = os.platform();
+    if (platform != "win32" && platform != "darwin")
+      return null;
+    const arch = os.arch();
+    if (arch != "x64")
+      return null;
+    if (platform == "win32")
+      return ["https://github.com/JCWasmx86/Swift-MesonLSP/releases/download/v2.0/Swift-MesonLSP-win64.zip",
```
tristan957, line 36:
```
const artifacts: { [key: string]: { name: string, hash: string } } = {
  "win32-x64": { name: "Swift-MesonLSP-win64.zip", hash: "fc0d930eb67525309291907c07df8655920cdc651f17f2320d09cc5271a23876" },
  "darwin-x64": { name: "Swift-MesonLSP-macos12.zip", hash: "c10999081cef8b56ea164d9a170470c268045254829e48bf4ac137a07a8a874c" },
};

const artifact = artifacts[`${platform}-${arch}`];
return [`${this.repoURL}/releases/download/v2.0/${artifact.name}`, artifact.hash];
```

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,47 @@
...
+    super("Swift-MesonLSP", ctx);
+  }
+
+  get downloadInfo(): [url: string, hash: string] | null {
+    const platform = os.platform();
+    if (platform != "win32" && platform != "darwin")
+      return null;
+    const arch = os.arch();
+    if (arch != "x64")
+      return null;
```
tristan957, line 34: There is duplicated logic here and in supportsSystem

src/lsp.ts
JCWasmx86: Will do that, do you have any linter/formatter for this repo that can do that automatically?

src/lsp/swift-mesonlsp.ts
tristan957: `artifacts` can be made a private member variable.

src/lsp.ts
JCWasmx86: I have difficulties finding a name. That's the workflow I've thought of (Pseudo code):


1. If the language server is installed, just start it and do nothingmore
2. If the OS is linux and a supported distribution+version: Download script that sets up the package sources and installs it
3. If the OS is linux and is not a supported distribution: Redirect user to the browser with the instructions to install from source
4. If the OS is Windows/MacOS: Just download+extract the zip files

A fallback could be just taking a nearly statically linked Ubuntu 18.04 binary and hoping for the best

src/lsp.ts
JCWasmx86: No, as e.g. the PATH can change (Drop binary into /usr/bin) and findLanguageServer() returns the path to the downloaded executable => Not cachable if I understand correctly

src/lsp.ts
JCWasmx86: I got this error:
```
src/extension.ts:292:26 - error TS2345: Argument of type 'SwiftMesonLspLanguageClient' is not assignable to parameter of type '{ dispose(): any; }'.
  Property 'dispose' is private in type 'SwiftMesonLspLanguageClient' but not in type '{ dispose(): any; }'.

292   ctx.subscriptions.push(client)
                             ~~~~~~


Found 1 error in src/extension.ts:292
```
After changing to private. Just having `dispose () {` seems to compile.

src/lsp.ts
tristan957: @lukester1975 has a PR to add prettier support which I think is a good idea. Then we can also work on adding eslint and various other tools as needed to aid in the development/review process.

src/lsp.ts
tristan957: Oh, I guess it is required by VSCode. Nevermind me :)

src/lsp.ts
tristan957, line 54: What is the second argument to this constructor?

JCWasmx86: It's this string here:

image

src/lsp.ts
tristan957: I see now why you were passing in serverModule into startLanguageServer. Hmm

src/lsp.ts
```diff
@@ -0,0 +1,180 @@
...
+    const lspDir = path.join(this.getExtensionLSPDir(this.context), this.executableName);
+    const suffix = platform == "win32" ? ".exe" : "";
+    const fullpath = path.join(lspDir, executableName + suffix);
+    if (fs.existsSync(fullpath)) {
+      return fullpath;
+    }
+    return null;
+  }
+
+  private getExtensionLSPDir(ctx: vscode.ExtensionContext): string {
```
tristan957, line 170: We don't need to take ctx here since you are storing it.

src/lsp.ts
tristan957: Oh. Should the whole Meson extension use the output channel? To me that would make sense. What are your thoughts?

src/lsp.ts
tristan957: Ok, I think we should just resolve the path in the constructor, and add a command to "restart the language server" similar to clangd. Then a user can change the path of the language server all they want.

src/lsp.ts
tristan957: But maybe that is not possible? If not, then this is fine :).

JCWasmx86: There is e.g. extra output from my language server that would e.g. help to diagnose issues and some other stuff

image


image

tristan957:
> Disabling muon when your language server is detected sounds good to me.

Logging a message to the output channel of the extension would be good in this case too.

src/lsp.ts
tristan957: Is there any way to just add a prefix to log messages coming from your LS?

JCWasmx86: I don't know, as I'm not that well-versed with the VSCode API

tristan957: Just leave as is, and we can explore this more later since it really isn't that important.

src/lsp.ts
```diff
@@ -0,0 +1,180 @@
...
+    );
+    fs.mkdirSync(
+      lspDir,
+      { recursive: true }
+    );
+    const tmpPath = path.join(os.tmpdir(), `lsp-${Date.now()}.zip`);
+    try {
+      this.downloadFile(downloadInfo[0], tmpPath).then(() => {
+        this.computeFileHash(tmpPath).then(str => {
+          const expected = downloadInfo[1];
```
tristan957, line 84: I think when you have named tuples, you can use `downloadInfo.hash`

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,47 @@
+import * as os from "os";
+
+import {
+  ExtensionContext,
+} from "vscode"
+import { LanguageServerClient } from "../lsp";
+import { Executable } from "vscode-languageclient/node";
+
+export class SwiftMesonLspLanguageClient extends LanguageServerClient {
+  private artifacts: { [key: string]: { name: string, hash: string } } = {
```
tristan957, line 10: static just for good measure.

tristan957:
Move the src/lsp.tx file to src/lsp/index.ts

tristan957:
I now have your branch pulled locally.

tristan957:
Might as well rebase on your next push. I think the only actual comment left at the moment is adding a "restart language server command". Then, it seems like we should restart the language server if "mesonbuild.languageServer" OR "mesonbuild.languageServerPath" change.

src/extension.ts
```diff
@@ -199,10 +198,54 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
+  } else if (canDownload && downloadLanguageServer == "yes") {
+    registerTooling = false;
+    client.setupLanguageServer();
+  }
+
+  if (registerTooling) {
+    activateLinters(root, ctx);
+    activateFormatters(ctx);
+  } else {
+    getOutputChannel().appendLine("Not enabling muon linter/formatter as language server is active!");
```
tristan957, line 248:
```suggestion
    getOutputChannel().appendLine("Not enabling muon linter/formatter as language server is active!");
    getOutputChannel().appendLine("Not enabling the muon linter/formatter because the the language server is active!");
```

src/lsp.ts
JCWasmx86: No, does not seem to work:

```

> mesonbuild@1.9.0 compile
> tsc -p ./

src/lsp/index.ts:84:41 - error TS2339: Property 'hash' does not exist on type '[url: string, hash: string]'.

84           const expected = downloadInfo.hash;
                                           ~~~~


Found 1 error in src/lsp/index.ts:84
```

tristan957: Interesting. Had no idea. I guess it is purely for documentation purposes then.

package.json
```diff
@@ -86,6 +86,10 @@
       {
         "command": "mesonbuild.benchmark",
         "title": "Meson: Run Benchmarks"
+      },
+      {
+        "command": "mesonbuild.restart",
```
tristan957, line 91: Maybe the command name should be "mesonbuild.restartLanguageServer"

src/extension.ts
```diff
@@ -248,4 +291,12 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
     }
   }
+
+  ctx.subscriptions.push(client);
+
+  ctx.subscriptions.push(
+    vscode.commands.registerCommand("mesonbuild.restart", async () => {
+      client.restart();
+    })
+  );
```
tristan957, line 301: What is `client` when a binary isn't found?

src/lsp/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      return;
+
+    this.startLanguageServer();
+  }
+
+  abstract supportsSystem(os: string, arch: string): boolean
+
+  abstract get downloadInfo(): [url: string, hash: string] | null;
+
+  restart(): void {
```
tristan957, line 39: In this function, we want to re-resolve the language server's path.

src/extension.ts
JCWasmx86: If no binary is found, it basically does nothing.

src/lsp/index.ts
JCWasmx86: It is doing this automatically, as it  is computed again by referencing `languageServerPath`, if I understand correctly

JCWasmx86:
I've added a temporary commit that shows another approach. A language server can be setup in two ways: Automatically and manually. For Windows/MacOS it is automatically, while Linux users are redirected to the documentation.

While I've not given up on the idea of basically doing `curl https://github.com/some/setup/script/for/linux.sh | sudo sh` to automatically setup the package sources for integration with the systems' package manager, I think there would be some hurdles (Would it work in confined environments like flatpak/snaps etc.)

So requiring manual setup for Linux users is probably inevitable unless you want to either download a binary for an ancient glibc (E.g. Ubuntu 18.04) and hope to the best. But AFAIK it would make problems for e.g. NixOS.  Or you use a script that will probably fail on some systems due to sandboxing or other issues.

package.json
```diff
@@ -86,6 +86,10 @@
       {
         "command": "mesonbuild.benchmark",
         "title": "Meson: Run Benchmarks"
+      },
+      {
+        "command": "mesonbuild.restartLanguageServer",
+        "title": "Meson: Restart language server"
```
tristan957, line 92: Title case. "Meson: Restart Language Server"

src/extension.ts
```diff
@@ -199,12 +199,81 @@ export async function activate(ctx: vscode.ExtensionContext) {
           break;
       }
     }
+  }
 
-    if (configureOnOpen === true) {
-      runFirstTask("reconfigure");
+  const downloadLanguageServerKey = "downloadLanguageServer";
```
tristan957, line 204: Can you create an enum for the settings instead of having these strings located throughout the code?

src/lsp/index.ts
```diff
@@ -0,0 +1,183 @@
...
+import * as vscode from "vscode";
+import { ExtensionContext, WorkspaceConfiguration, workspace } from "vscode";
+import { Executable, LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } from "vscode-languageclient/node";
+
+export abstract class LanguageServerClient {
+  private config: WorkspaceConfiguration;
+  private ls: LanguageClient | null = null;
+  executableName: string;
+  private context: ExtensionContext;
+  private clientOptions: LanguageClientOptions = {
```
tristan957, line 17: This should be private static

src/lsp/index.ts
```diff
@@ -0,0 +1,183 @@
...
+import * as os from "os";
+import * as path from "path";
+import * as vscode from "vscode";
+import { ExtensionContext, WorkspaceConfiguration, workspace } from "vscode";
+import { Executable, LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } from "vscode-languageclient/node";
+
+export abstract class LanguageServerClient {
+  private config: WorkspaceConfiguration;
+  private ls: LanguageClient | null = null;
+  executableName: string;
```
tristan957, line 15: Does this need to be abstract?

src/lsp/index.ts
tristan957: You are correct on this aspect, but I think it would be best to resolve the path in the constructor as mentioned in a previous comment.

src/extension.ts
JCWasmx86: Do you mean like this: https://github.com/mesonbuild/vscode-meson/pull/123/commits/dbb15f11b248687894976efa299ba23904775c84#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR133

tristan957:
I have been hacking on your branch. I will submit a PR to your remote.

src/lsp/index.ts
JCWasmx86: I don't understand what should be resolved in the constructor. Resolving `which.sync(this.executableName, { nothrow: true })` doesn't make sense, as there is a workflow like:

- Linux user starts VSCode
- Installs LS into /usr/local/bin manually
- Restarts the language server

And because this part is resolved earlier, the language server would still fail to change.

`this.findLanguageServer()` doesn't make sense too:
- Windows/MacOS user starts VSCode
- LS is installed
- `restart()`
- The already resolved result from `findLanguageServer()` is still null

And for `this.config.languageServerPath` I don't see a use case, too.
- User changes config, switches to some other language server
- `restart()`
- The already resolved value would still be the old language server

So what exactly do you mean?

package.json
JCWasmx86: This thread will be obsoleted by https://github.com/JCWasmx86/vscode-meson/pull/3

JCWasmx86:
I'm not sure how I should restructure the commits, as there are e.g. temporary commits contained within and some by @tristan957. I would just make it one commit `Add integration for Swift-MesonLSP` and manually update the authorship

tristan957:
Squash it all down assuming you're happy. Just add me with a coauthor trailer.

tristan957:
I'll take another look at the PR in the next couple of days.

JCWasmx86:
Oops, messed up the commit history of prior commits

JCWasmx86:
I'm available again for addressing reviews. Could you please review especially the last commit? `tmp: Controversial changes.`. 

Either it's my somewhat mediocre TS+async model skills, or there was something wrong with the code before (As I probably did some mistake during testing).


- With the hashing I had problems due to `reject`ing for some unknown reason. I fixed it by moving everything into the returned promised and removing the `pipe` to the hash-object.
- I switched from a promise-based architecture to a "normal" architecture, as I couldn't get changes to the local variable `uri` to be returned. Furthermore the code is more clear now, ymmv.

JCWasmx86:
Thinking more about this, there should be some way to disable the language server for a project. There are at least three projects that use meson that would get *a lot* of errors due to the project structure:

- [HelenOS](http://www.helenos.org/)
- [picolibc](https://github.com/picolibc/picolibc)
- [DPDK](https://www.dpdk.org/)

All of these stem from the fact that I haven't implemented an interpreter, instead I'm basically guessing the parameters to `set_variable` and `subdir` using some [weird mix of heuristics and partially interpreting](https://github.com/JCWasmx86/Swift-MesonLSP/blob/main/Sources/MesonAnalyze/SetVariableGuess.swift)

HelenOS is basically not possible for my language server, as it has a config system that generates parts of the build AFAIU. 

Both picolibc and DPDK have a code structure that is more difficult for my partial interpreter to understand

tristan957:
That makes sense to me. Wanna add an option for it?

JCWasmx86:
Can do that

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## next
 
+- Add integration for [Swift-MesonLSP](https://github.com/JCWasmx86/Swift-MesonLSP).
```
tristan957, line 5: While you are rebasing, s/for/with

package.json
```diff
@@ -187,6 +191,35 @@
             ]
           },
           "description": "Give an object that's merged into the debug configuration provider"
+        },
+        "mesonbuild.languageServer": {
+          "type": "string",
+          "default": "Swift-MesonLSP",
+          "enum": [
+            "Swift-MesonLSP",
+            "null"
```
tristan957, line 200: Can we use the null keyword in JSON instead here?

package.json
```diff
@@ -187,6 +191,35 @@
...
+          "default": "ask",
+          "enum": [
+            true,
+            false,
+            "ask"
+          ],
+          "enumDescriptions": [
+            "Download the language server",
+            "Do not download the language server",
+            "Ask every time"
```
tristan957, line 220: Might be worth finishing this string with `to download the language server`

src/extension.ts
```diff
@@ -186,25 +184,80 @@ export async function activate(ctx: vscode.ExtensionContext) {
           break;
 
         case Options.never:
-          extensionConfigurationSet(configureOnOpenKey, false, vscode.ConfigurationTarget.Workspace);
+          extensionConfigurationSet(SettingsKey.ConfigureOnOpen, false, vscode.ConfigurationTarget.Workspace);
```
tristan957, line 187: I would like the casing of `ConfigureOnOpen` to match `configureOnOpen`. Similar for the other keys

src/extension.ts
```diff
@@ -186,25 +184,80 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
+
+      case Options.no:
+        extensionConfigurationSet(SettingsKey.DownloadLanguageServer, "ask", vscode.ConfigurationTarget.Global);
+        return false;
     }
+
+    return false;
   }
 
+  let client = await createLanguageServerClient(server, await shouldDownload(downloadLanguageServer), ctx);
```
tristan957, line 236: It looks like this could be const.

src/extension.ts
```diff
@@ -186,25 +184,80 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
+  if (client !== null) {
+    ctx.subscriptions.push(client);
+    client.start();
+
+    getOutputChannel().appendLine("Not enabling the muon linter/formatter because the language server is active.");
```
tristan957, line 241: Only output this line if the language server is yours (granted it is the only one haha). Also explicitly add Swift-MesonLsp to the output line.

src/extension.ts
```diff
@@ -186,25 +184,80 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
+    activateFormatters(ctx);
+  }
+
+  ctx.subscriptions.push(
+    vscode.commands.registerCommand("mesonbuild.restartLanguageServer", async () => {
+      if (client === null) {
+        client = await createLanguageServerClient(server, await shouldDownload(downloadLanguageServer), ctx);
+        if (client !== null) {
+          ctx.subscriptions.push(client);
+          client.start();
```
tristan957, line 253: The output line from above about not enabling muon would be good to have here, though I am not sure of the best factoring.

src/lsp/common.ts
```diff
@@ -0,0 +1,53 @@
+import * as vscode from "vscode";
+import { LanguageServerClient } from ".";
+import { LanguageServer } from "../types";
+import { SwiftMesonLspLanguageClient } from "./swift-mesonlsp";
+import { Uri } from "vscode";
+
+export async function createLanguageServerClient(server: LanguageServer, download: boolean, context: vscode.ExtensionContext): Promise<LanguageServerClient | null> {
```
tristan957, line 7: I think this might make sense to factor into the LanguageServerClient class.

```typescript
public static async LanguageServerClient.create(...): Promise<LanguageServerClient | null>;
```

src/lsp/swift-mesonlsp.ts
```diff
@@ -0,0 +1,64 @@
...
+import * as vscode from "vscode";
+
+import { LanguageServerClient } from "../lsp";
+import { Executable } from "vscode-languageclient/node";
+
+export class SwiftMesonLspLanguageClient extends LanguageServerClient {
+  private static artifacts: { [key: string]: { name: string, hash: string } } = {
+    "win32-x64": { name: "Swift-MesonLSP-win64.zip", hash: "1bb06cec242251bcf7ea58e173a46105039fbdf381c3c0ab6edd3b7c3798dbfa" },
+    "darwin-x64": { name: "Swift-MesonLSP-macos12.zip", hash: "e016c9041f57ed452ff8d45fb2a28e82dd446d7606bf18107b572b3cbd82fd20" },
+    "linux-x64": { name: "Swift-MesonLSP.zip", hash: "0a8d1eecccbc35d4b3d49444824a44c5da2430a34692f706ba34acaf12322efa" }
```
tristan957, line 11: If you have had any new releases, might as well update the refs

tristan957 (commented):
Just these last few comments!

src/extension.ts
JCWasmx86: No, as a few lines later I do:
```ts
  ctx.subscriptions.push(
    vscode.commands.registerCommand("mesonbuild.restartLanguageServer", async () => {
      if (client === null) {
        client = await createLanguageServerClient(server, await shouldDownload(downloadLanguageServer), ctx); // <<- HERE
        if (client !== null) {
          ctx.subscriptions.push(client);
          client.start();
        }
      } else {
        client.restart();
      }
    })
  );
```

src/extension.ts
JCWasmx86: Wouldn't we have to stop linting using muon there? And as far as I can see, there's currently no way to do that

tristan957: Hmm good point. Wanna add a TODO on that front for if we ever figure it out?

src/lsp/common.ts
JCWasmx86: Sadly moving it does not work:
```
2023-08-08 18:45:28.158 [error] TypeError: Class extends value undefined is not a constructor or null
	at Object.<anonymous> (/home/user/Projects/vscode-meson/out/lsp/swift-mesonlsp.js:6:49)
	at Module._compile (node:internal/modules/cjs/loader:1141:14)
	at Module._extensions..js (node:internal/modules/cjs/loader:1196:10)
	at Module.load (node:internal/modules/cjs/loader:1011:32)
	at Module._load (node:internal/modules/cjs/loader:846:12)
	at f._load (node:electron/js2c/asar_bundle:2:13330)
	at o._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:34104)
	at f._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:30479)
	at I._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:94:24270)
	at Module.require (node:internal/modules/cjs/loader:1035:19)
	at require (node:internal/modules/cjs/helpers:102:18)
	at Object.<anonymous> (/home/user/Projects/vscode-meson/out/lsp/index.js:15:26)
	at Module._compile (node:internal/modules/cjs/loader:1141:14)
	at Module._extensions..js (node:internal/modules/cjs/loader:1196:10)
	at Module.load (node:internal/modules/cjs/loader:1011:32)
	at Module._load (node:internal/modules/cjs/loader:846:12)
	at f._load (node:electron/js2c/asar_bundle:2:13330)
	at o._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:34104)
	at f._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:30479)
	at I._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:94:24270)
	at Module.require (node:internal/modules/cjs/loader:1035:19)
	at require (node:internal/modules/cjs/helpers:102:18)
	at Object.<anonymous> (/home/user/Projects/vscode-meson/out/lsp/common.js:5:12)
	at Module._compile (node:internal/modules/cjs/loader:1141:14)
	at Module._extensions..js (node:internal/modules/cjs/loader:1196:10)
	at Module.load (node:internal/modules/cjs/loader:1011:32)
	at Module._load (node:internal/modules/cjs/loader:846:12)
	at f._load (node:electron/js2c/asar_bundle:2:13330)
	at o._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:34104)
	at f._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:30479)
	at I._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:94:24270)
	at Module.require (node:internal/modules/cjs/loader:1035:19)
	at require (node:internal/modules/cjs/helpers:102:18)
	at Object.<anonymous> (/home/user/Projects/vscode-meson/out/extension.js:13:18)
	at Module._compile (node:internal/modules/cjs/loader:1141:14)
	at Module._extensions..js (node:internal/modules/cjs/loader:1196:10)
	at Module.load (node:internal/modules/cjs/loader:1011:32)
	at Module._load (node:internal/modules/cjs/loader:846:12)
	at f._load (node:electron/js2c/asar_bundle:2:13330)
	at o._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:34104)
	at f._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:30479)
	at I._load (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:94:24270)
	at Module.require (node:internal/modules/cjs/loader:1035:19)
	at require (node:internal/modules/cjs/helpers:102:18)
	at Function.i [as __$__nodeRequire] (/usr/share/codium/resources/app/out/vs/loader.js:5:98)
	at r._loadCommonJSModule (/usr/share/codium/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:130:32106)
	at async Promise.all (index 0)
```
That was the reason in the beginning I moved it into the common.ts. The explanation according to SO was some cyclic import/dependency cycle or something like that.

src/extension.ts
JCWasmx86: Yes, will do that

src/lsp/common.ts
tristan957: I'll pull the latest branch this week to try it for myself just to see what I can dig up

JCWasmx86:
For some reason the download/hashing/extracting is horrendously broken. (I thought I fixed it a while ago, but not sure why it exactly fails)

It seems to be easy:

- Download
- Hash
- Extract

But for some reason it does:
- Download + Hash concurrently, so the hash ends up being an empty string => Fail

Have to debug that a bit more.

tristan957:
Sounds good. I am on Linux with your COPR enabled, so I should be able to test the cyclic import thing on my end :)

JCWasmx86:
Sadly I still wasn't able to figure out, why it attempts to hash an empty string instead of the contents, maybe you have a better luck with finding the cause for it

JCWasmx86:
I got it working now. For some reason github returned 302 instead of 300/301, which I wasn't able to handle => This PR is ready for maybe the final review?

tristan957:
Cool wanna squash the commits where appropriate. I wanna try to get to it this upcoming week. Thanks for your continued work.

JCWasmx86:
```
Run tsc --noEmit --noErrorTruncation --pretty false
Error: src/utils.ts(31,33): error TS2322: Type 'ExecFileException | undefined' is not assignable to type 'ExecException | undefined'.
  Type 'ExecFileException' is not assignable to type 'ExecException | undefined'.
    Type 'ExecFileException' is not assignable to type 'ExecException'.
      Types of property 'code' are incompatible.
        Type 'string | number | null | undefined' is not assignable to type 'number | undefined'.
          Type 'null' is not assignable to type 'number | undefined'.
Error: Process completed with exit code 2.
```
Not sure why exactly it breaks, but I broke it

tristan957:
Did you rebase on master?

JCWasmx86:
Yes, already reformatted everything

JCWasmx86:
I've accidentally bumped the `@types/node` version during fixing conflicts, but it's fixed now

tristan957:
Ha. Thanks for rebasing :).

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
 - Add syntax highlight for `meson.options` files.
 - Fix listing tests in the Test Explorer on activation
 - Add support to run executables that are not installed (eg: examples)
+- Add integration with
+  [Swift-MesonLSP](https://github.com/JCWasmx86/Swift-MesonLSP). vscode-meson
+  will automatically download it on Windows, Linux and MacOS. Linux users can
+  follow the [docs](https://github.com/JCWasmx86/Swift-MesonLSP/tree/main/Docs)
+  if they want to have updates delivered using their package manager.
```
tristan957, line 14: Move this to the next section

package.json
```diff
@@ -338,14 +371,20 @@
...
+    "@types/which": "^3.0.0",
     "husky": "^8.0.3",
     "lint-staged": "^14.0.1",
     "prettier": "^3.0.3",
     "typescript": "^4.4.4"
   },
-  "dependencies": {},
+  "dependencies": {
+    "adm-zip": "0.5.10",
+    "vscode-languageclient": "8.0.2",
```
tristan957, line 385: Mind updating vscode-languageclient? Let's add `^` to these versions

PR #125, opened 2023-04-13
Add a new command to run executables

ylatuya:
This adds a new command named `Meson: Run Executable` that lets the user select from a list of all the executables targets and run one of them, similar to how `Meson: Run Unit Tests` works.

CC @xclaesse

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -7,7 +7,7 @@ import {
   runMesonReconfigure,
   runMesonInstall
 } from "./meson/runners";
-import { getMesonTasks } from "./tasks";
+import { getMesonTasks, getTask, getTasks } from "./tasks";
 import { MesonProjectExplorer } from "./treeview";
 import { TargetNode } from "./treeview/nodes/targets"
 import {
@@ -36,6 +36,7 @@ import {
 import {
   activateFormatters
 } from "./formatters"
+import { TaskQuickPickItem } from "./types";
 
 export let extensionPath: string;
 let explorer: MesonProjectExplorer;
@@ -186,6 +187,12 @@ export async function activate(ctx: vscode.ExtensionContext) {
     })
   );
 
+  ctx.subscriptions.push(
+    vscode.commands.registerCommand("mesonbuild.run", async () => {
+      runExecutable();
+    })
+  );
+
   const configureOnOpenKey = "configureOnOpen";
   let configureOnOpen = extensionConfiguration(configureOnOpenKey);
   if (configureOnOpen === "ask") {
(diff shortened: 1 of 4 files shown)


xclaesse:
OOC, how does it work to get that `Meson: Run Executable` task? In the `Run tasks...` menu I only see individual executables. I don't get a picker menu.

I think we need to fix executable tasks a bit more:
- Is it possible to pass them the envFile?
- We should at least exclude unit test executables. I think we should use the same heuristic that devenv: only include executables that are being installed in bindir.

ylatuya:
I didn't see that all executables have already a task that can be launched from the `Run tasks...` menu. I still think this could be useful to split between tests and executables, as right now all of them are grouped as `vscode.TaskGroup.Test` so that new menu will allow launching executables that are not tests.

A possible improvement would be:
-  Keep tests in the `vscode.TaskGroup.Test`
-  Create a new group for executables that are not tests, using the filtering logic you proposed
- Add the env parameter to vscode.ProcessExecution to that executable are run with the correct environment

src/extension.ts
```diff
@@ -308,4 +315,46 @@ export async function activate(ctx: vscode.ExtensionContext) {
 
     explorer.refresh();
   }
+
+  async function pickExecutable() {
```
xclaesse, line 319: This function is much better than pickTestOrBenchmark() above, would be nice if we could have some common code. After all they both select a task, it's just the mode that change.

src/extension.ts
```diff
@@ -308,4 +315,46 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
+    let taskItem;
+    try {
+      taskItem = await pickExecutable();
+    } catch (err) {
+      // Pick cancelled.
+    }
+    if (taskItem != null) {
+      await vscode.tasks.executeTask(taskItem.task);
+    }
+    explorer.refresh();
```
xclaesse, line 358: I don't think that's needed, I noticed when I worked on my previous patch that it's really costly to call that, I removed most of them already. Only needed in changeHandler AFAIK?

src/extension.ts
xclaesse: I think we can do a massive simplification: https://github.com/mesonbuild/vscode-meson/commit/13a9fd0433062c5be2afaa776be1113c913104b0

xclaesse:
> I didn't see that all executables have already a task that can be launched from the `Run tasks...` menu.

This PR actually relies on that, all it does is picking a Task that already exist.

> I still think this could be useful to split between tests and executables, as right now all of them are grouped as `vscode.TaskGroup.Test` so that new menu will allow launching executables that are not tests.

I don't know what are the possible values for TaskGroup, but that should be tweaked indeed.

>     * Add the env parameter to vscode.ProcessExecution to that executable are run with the correct environment

I changed that already, it got merged.

src/extension.ts
xclaesse: I refactored that in https://github.com/mesonbuild/vscode-meson/pull/127, I don't think it is a blocker to get this PR merged first.

PR #127, opened 2023-04-14
Misc improvements

xclaesse:
- Remove src/meson directory
    
    It does not contain much, better merge what's remaining into toplevel
    src/ dir. The split is a bit arbitrary anyway, since the whole extension
    is about meson.

- Merge configure and reconfigure tasks
    
    Meson >=1.1.0 already accepts passing --reconfigure regardless whether
    the builddir is configured already or not. Simulate the same behaviour
    by merging both tasks and add --reconfigure only if needed.
    
    This also run the initial configuration in the terminal instead of
    hiding output.

- Do not set env when running tests
    
    We run "meson test" command which already sets the proper env. We are
    not running the test executable directly.

- Remove default buildtype option
    
    Most projects already have it in their default_options. If not, "debug"
    is already Meson's default.

- Implement all commands by running its corresponding task

- Refactor generic pickAndRunTask() function


dcbaker (approved):

dabrain34:
Thanks @xclaesse that will be a nice improvement !

PR #128, opened 2023-04-14
env file might have \r\n newline

xclaesse:


src/utils.ts
```diff
@@ -188,10 +188,10 @@ export async function genEnvFile(buildDir: string) {
   const data = await fs.promises.readFile(envfile);
   for (const i of data.toString().split("\n")) {
```
dcbaker, line 189: I think we could just make the split more robust here using a regex literal:
```typescript
for (const i of data.toString().split(/\r?\n/)) {
    ...
}
```
And then we can just work with the string as-is without the need to call trimEnd

dcbaker (changes requested):

src/utils.ts
xclaesse: That works too, not sure which one is best tbh. Regex are often more complicated to read and easier to get wrong.

dcbaker: I guess we could also change meson to always output a `\n` only file?

I think I prefer the regex approach because it means that the loop body is easier to reason about, but I could go either way

xclaesse: I changed to use your regex. It could be fixed in meson indeed, but that would require using newer version of meson, better keep working with 1.1.0 that just got released.

PR #131, opened 2023-04-16
Only run linter on meson files

annacrombie:


jkrenzer:
Just wanted to ask, if you could get this patch into a new release, so vscode could install the patched version of the extension. The most current v1.9.0 seems to be from 13-04-2023 and thus the linter is still unusable.

Thank!

tristan957:
@dcbaker thoughts?

PR #132, opened 2023-04-18
Fix version showing "undefined" in tree view

tristan957:
Meson does not require a version to be set in the project() call. When a version is not set, Meson sets the version as "undefined" in the introspection data.

### DIFF
--- a/src/treeview/nodes/toplevel.ts
+++ b/src/treeview/nodes/toplevel.ts
@@ -21,7 +21,9 @@ export class ProjectNode extends BaseNode {
   getTreeItem() {
     const item = super.getTreeItem() as vscode.TreeItem;
 
-    item.label = `${this.project.descriptive_name} ${this.project.version}`;
+    item.label = `${this.project.descriptive_name}`;
+    if (this.project.version !== "undefined")
+      item.label += ` (${this.project.version})`;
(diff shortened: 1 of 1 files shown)


tristan957:
I have no clue why the introspection data doesn't use `null`. Feels like a huge bug to me.

@dcbaker might want to require 1 approval for PRs. Currently I could merge my own PR if I wanted.

src/treeview/nodes/toplevel.ts
```diff
@@ -99,7 +101,7 @@ class SubprojectNode extends BaseNode {
     benchmarks: Tests,
 
     ) {
-    super(`${parentId}-${subproject.descriptive_name}-${subproject.version}`);
+    super(`${parentId}-${subproject.descriptive_name}${subproject.version !== "undefined" ? `-${subproject.version}` : ""}`);
```
xclaesse, line 104: Why are you doing it in a single line here and not in other places?

tristan957: JavaScript has a restriction that `super()` has to be the first statement in a constructor. Otherwise, I would have moved it too. Java has a similar problem.

xclaesse: Ok lgtm

PR #134, opened 2023-04-19
extension: refresh / clear cache if extension configuration is changed.

lukester1975:
Otherwise there's staleness since ce4caf62a1610b83748a689e167e4223677baa50. I just hooked the existing `changeHandler`. Seems OK, though obviously doesn't work for builddir changes...

Speaking of which, is there any plan to implement something similar to #103? I'll have to rebase it for my own needs, if not.

Thanks

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -53,8 +53,8 @@ export async function activate(ctx: vscode.ExtensionContext) {
       vscode.DebugConfigurationProviderTriggerKind.Dynamic)
   );
 
-  let updateHasProject = async () => {
-    let mesonFiles = await vscode.workspace.findFiles("**/meson.build");
+  const updateHasProject = async () => {
+    const mesonFiles = await vscode.workspace.findFiles("**/meson.build");
     vscode.commands.executeCommand("setContext", 'mesonbuild.hasProject', mesonFiles.length > 0);
   }
   mesonWatcher = vscode.workspace.createFileSystemWatcher("**/meson.build", false, true, false);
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -68,23 +68,21 @@ export async function activate(ctx: vscode.ExtensionContext) {
   controller.createRunProfile("Meson run test", vscode.TestRunProfileKind.Run, (request, token) => testRunHandler(controller, request, token), true)
   ctx.subscriptions.push(controller);
 
-  let mesonTasks: Thenable<vscode.Task[]> | undefined = undefined;
+  let mesonTasks: Thenable<vscode.Task[]> | null = null;
```
xclaesse, line 71: OOC, what difference does it make to use null instead of undefined? I copied that line verbatim from vscode doc...

src/extension.ts
```diff
@@ -96,7 +94,16 @@ export async function activate(ctx: vscode.ExtensionContext) {
   ctx.subscriptions.push(watcher);
   await genEnvFile(buildDir);
 
-  let compileCommandsHandler = async () => {
+  // Refresh if the extension configuration is changed. builddir changes won't be reflected, however.
```
xclaesse, line 97: What would it take to reflect builddir change? That seems to be the most important option that should trigger a refresh.

xclaesse: Can be done in separate PR if we have a way of course, not blocking this PR.

src/extension.ts
lukester1975: Sorry I didn't really mean to commit that, was going to suggest some cleanups separately (after turning on `strictNullChecks`, but didn't really have to time to finish the job).

More because of the assigning `undefined` later which I think is considered unpleasant. Crazy language design!

src/extension.ts
lukester1975: Well, #103 😀 I am not looking forward to rebasing it...

Maybe for now the simplest option is to reload the whole extension, given how early on `buildDir` is read and cached after activation?

xclaesse: > Maybe for now the simplest option is to reload the whole extension, given how early on buildDir is read and cached after activation?

Fair enough, can you open a PR please?

PR #136, opened 2023-04-19
Instead of target type in ui, use a tool tip.

lukester1975:
Just a suggestion. I understand the need for c5a5166 but IMHO it clutters the UI a tad.

Since there's already separate icons for lib/exe etc., why not distinguish static and shared libs that way? Here's an attempt. It's dynamic because it's very slightly slanted, obviously 😀

Then use tooltip for the text form.


tristan957:
I like this change. If someone else agrees, let's merge it.

tristan957 (approved):

src/treeview/nodes/targets.ts
```diff
@@ -103,7 +103,8 @@ export class TargetNode extends BaseNode {
   async getTreeItem() {
     const item = super.getTreeItem() as vscode.TreeItem;
 
-    item.label = `${this.target.name} ${this.target.type}`;
+    item.label = this.target.name;
+    item.tooltip = `${this.target.name} (${this.target.type})`;
```
xclaesse, line 107: Is it worth repeating the target name in the tooltip?

tristan957: I don't think so.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Automatically configure Intellisense to use compile_commands.json generated
   by Meson.
+- Meson `tasks.json` entries specifying `target` will now need to include the target's type. E.g. `"target": "relative/path/to/targetname:shared_library"`.
```
tristan957, line 9: I admit I am not an expert of tasks.json, but I hate magic strings like this.

Can we do something like the following?

`
{
  "target": { "path": "rel/path", "type" : "the type which is an enum" }
}

lukester1975: IMHO since it's just the meson command line target name parameter, but with a mandatory type, personally I prefer it as the single string even though I agree with your principle. Plus there would be changes to the task generation and node commands presumably.

Sorry, not mine it was from c5a5166, just thought I'd add the changelog note whilst doing related stuff as it took a while to figure out why my tasks were broken...

PR #137, opened 2023-04-19
extension: reload window if buildFolder changes.

lukester1975:
Following on from #134

FWIW. Very sledgehammer / nut. I can't see a trivial and more fine-grained approach (e.g. just deactivate/reactivate extension) and given I want to rebase code that already decouples builddir, I don't really want to look any further. 

That said I did see this: https://github.com/microsoft/vscode/issues/45774#issuecomment-373423895

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         "mesonbuild.buildFolder": {
           "type": "string",
           "default": "builddir",
-          "description": "Specify in which folder Meson build configure and build the project."
+          "description": "Specify in which folder Meson build configure and build the project. Changing this value will reload the VS Code window."
         },
         "mesonbuild.configureOptions": {
           "type": "array",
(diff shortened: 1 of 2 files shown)


tristan957:
Did you try the other thing in the comment?

lukester1975:
Nope, like I say I already have a branch that supports multiple builddirs, so whatever time I get I'll spend on getting that back up to speed.

tristan957:
Ok. So then this is kind of a stop-gap until the other PR gets work done on it?

xclaesse:
Better have something that works first, then see how we can improve later. LGTM.

lukester1975:
> Ok. So then this is kind of a stop-gap until the other PR gets work done on it?

Mine? I wouldn't like to say how or if that would ever be merge-worthy, let alone if you guys like the approach. I've just done what I needed to work... Maybe I'll try to rebase it at the weekend if I have sufficient caffeine.

PR #138, opened 2023-04-19
Add test and benchmark options

lukester1975:
Being able to run tests with `--verbose` is sometimes useful.

Maybe `testOptions` (and a separate `benchmarkOptions`) is a better more generic approach?

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,18 @@
           "description": "Specify the list of options used for Meson setup. Can be used to specify a cross file (use --cross-file=FILE).",
           "deprecationMessage": "--cross-file and --native-file should be in configureOptions too."
         },
+        "mesonbuild.testOptions": {
+          "type": "array",
+          "default": [],
+          "description": "Specify the list of options used for running tests."
+        },
+        "mesonbuild.benchmarkOptions": {
+          "type": "array",
+          "default": [
+            "--verbose"
+          ],
+          "description": "Specify the list of options used for running benchmarks. --benchmark is implicit."
+        },
         "mesonbuild.mesonPath": {
           "type": "string",
           "default": "meson",
@@ -304,7 +316,7 @@
         "source": "gcc",
         "owner": "meson",
         "fileLocation": [
-          "autodetect",
+          "autoDetect",
           "${workspaceFolder}/${config:mesonbuild.buildFolder}"
         ],
         "pattern": {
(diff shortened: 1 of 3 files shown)


src/tasks.ts
```diff
@@ -18,9 +18,11 @@ interface MesonTaskDefinition extends vscode.TaskDefinition {
 function createTestTask(t: Test, buildDir: string, isBenchmark: boolean) {
   const project = t.suite[0].split(":")[0]
   const name = `${project}:${t.name}`;
-  const mode = isBenchmark ? "benchmark" : 'test'
-  const benchmarkArgs = isBenchmark ? ["--benchmark", "--verbose"] : [];
-  const args = ["test", ...benchmarkArgs].concat(name);
+  const mode = isBenchmark ? "benchmark" : "test";
+  const verboseArgs = (isBenchmark || extensionConfiguration("verboseTest")) ? ["--verbose"] : [];
```
xclaesse, line 22: Something I've been wondering before, why are benchmark verbose by default? Maybe we should remove this different now that we have an option?

package.json
```diff
@@ -175,6 +175,11 @@
             ]
           },
           "description": "Give an object that's merged into the debug configuration provider"
+        },
+        "mesonbuild.verboseTest": {
+          "type": "boolean",
+          "default": false,
+          "description": "Enable verbose test output"
```
xclaesse, line 182: If I understand correctly, it's not only tests, but build too?

tristan957:
testOptions and benchmarkOptions would be much preferred I think

src/tasks.ts
lukester1975: That was from me in #48. Didn't see the point of running a benchmark and not looking at the output :)

lukester1975:
OK. Separate strings and `benchmarkOptions` including `--verbose` by default sounds good to me, I'll take a look at some point.

tristan957:
I think making "mesonbuild.benchmarkOptions" default to `["--verbose"]` instead of doing any magic in the logic of the extension will be best. Hopefully that makes sense.

lukester1975:
Yeah, that's what I was getting at too.

lukester1975:
One more thought: #48 pre-dated https://github.com/mesonbuild/meson/commit/9a1a5c2b74627d8d0476def0ae11848308994b5f so made more sense / was required back then.

Anyway, here's the `testOptions` and `benchmarkOptions`.

tristan957 (approved):

PR #139, opened 2023-04-21
Added prettier, lint-staged and husky.

lukester1975:
As suggested in #68.

Just a stab at a prettier config. (I personally don't understand the preference for 80 columns... my monitor is wider than I am tall, and I'm 6'5")

print-width 100 example: https://github.com/lukester1975/vscode-meson/compare/78052ee5f2c33fcdaa25b93698726b10d658a71d..1fbd08082f804f73366ba032005d3101005d9f1d
80: https://github.com/lukester1975/vscode-meson/compare/78052ee5f2c33fcdaa25b93698726b10d658a71d..856fbf0fa785f62117d4af8da0ad58baa0fc2645


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
tristan957, line 1: `/bin/sh` is the preferred shebang I believe

.prettierignore
```diff
@@ -0,0 +1,3 @@
+.git
+node_modules
+out
```
tristan957, line 3:
```ignore
/.git/
/node_modules/
/out/
```

package.json
```diff
@@ -319,14 +319,21 @@
...
     "@types/vscode": "^1.1.59",
+    "husky": "^8.0.3",
+    "lint-staged": "^13.2.1",
+    "prettier": "^2.8.7",
     "typescript": "^4.4.4"
   },
-  "dependencies": {}
+  "dependencies": {},
+  "prettier": {
+    "printWidth": 100
```
tristan957, line 337: Let's just finish the rest of the config.

Determine if we are using tabs or spaces, and choose it.

semi: true
singleQuote: false
quoteProps: as-needed
jsxSingleQuote: false
trailingCommas: all
bracketSpacing: true
bracketSameLine: false
arrowParens: true
wrapProse: always
endOfLine: lf

tristan957 (commented):
Can you add a GitHub Action for using Prettier?

Here is an example: https://github.com/tristan957/libmerr/blob/master/.github/workflows/prettier.yaml

.husky/pre-commit
lukester1975: I wouldn't want to be on that jury...! However, husky creates the hooks like this (i.e. with `husky add pre-commit "yarn --silent lint-staged -r"`).

lukester1975:
GH action - sure, will take a look at some point.

dcbaker:
I haven't looked at anything but the initial MR message. I'd be happy with 100 or 120 column widths. I'd prefer not to go wider than that as I usually have two files opened side-by side. I think Meson itself is 100?

package.json
dcbaker: spaces

tristan957:
@lukester1975 any time to come back to this?

lukester1975:
I'm not a GH actions person but this appears to work based of yours @tristan957.

* meson uses a line length of 120 by the looks of it (in `.flake8`), so did the same here.
* Since there's an `.editorconfig`, I put the line length there as prettier will use that.

Not sure I see the point of specifying options with the default values: as soon as a new value is added, it's no longer complete anyway. Plus if a default ever changes, highlighting that fact and reviewing what is preferred is a good thing IMHO. Plus package.json is too long as it is...

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+# SPDX-FileCopyrightText: 2023 Tristan Partin <tristan@partin.io>
```
tristan957, line 3: TBH, you can just drop this. It isn't important to me

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,39 @@
...
+
+name: Prettier
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - README.md
```
tristan957, line 11:
```
- **.json
- **.md
- **.ts
- **.yml
```

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: latest
+
+      - name: Checkout
+        uses: actions/checkout@v3
```
tristan957, line 31: You can bump this to v4.

.github/workflows/prettier.yml
lukester1975: Heh, silly me.

That said, given the `prettier:check` script specifies the patterns, might as well not filter here? Not like PRs are flying in so even if there's a perf hit surely not an issue...

.github/workflows/codeql-analysis.yml
```diff
@@ -16,12 +16,12 @@ permissions:
```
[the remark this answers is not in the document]
lukester1975: Already is one apparently, so I've added the commit. Will try to remember it on next rebase, unless this is getting merged soon...

.github/workflows/prettier.yml
tristan957: That is fine. Can always change it later.

.github/workflows/codeql-analysis.yml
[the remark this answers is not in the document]
tristan957: I am ready to merge. Are you good to go?

tristan957:
Weird. I will watch it.

PR #140, opened 2023-04-23
Enable strict mode and fix up all resulting errors and warnings.

lukester1975:
This is the current tsconfig from https://github.com/Microsoft/vscode-generator-code with additional options:

```
noImplicitReturns
noFallthroughCasesInSwitch
noImplicitOverride
noPropertyAccessFromIndexSignature
```

Thanks

### DIFF
--- a/src/tasks.ts
+++ b/src/tasks.ts
@@ -70,7 +70,7 @@ function createReconfigureTask(buildDir: string) {
   );
 }
 
-export async function getMesonTasks(buildDir: string): Promise<vscode.Task[]> {
+export async function getMesonTasks(buildDir: string) {
   try {
     const defaultBuildTask = new vscode.Task(
       { type: "meson", mode: "build" },
@@ -119,8 +119,9 @@ export async function getMesonTasks(buildDir: string): Promise<vscode.Task[]> {
     ];
 
     // Remaining tasks needs a valid configuration
-    if (!checkMesonIsConfigured(buildDir))
+    if (!checkMesonIsConfigured(buildDir)) {
       return tasks;
+    }
 
     const [targets, tests, benchmarks] = await Promise.all([
       getMesonTargets(buildDir),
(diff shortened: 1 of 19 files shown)


src/tasks.ts
```diff
@@ -173,27 +179,36 @@ export async function getMesonTasks(buildDir: string): Promise<vscode.Task[]> {
...
+    (t) => t.definition["mode"] === mode
   );
 }
 
 export async function runTask(task: vscode.Task) {
   try {
     await vscode.tasks.executeTask(task);
-  } catch (e) {
-    vscode.window.showErrorMessage(`Could not ${task.definition.mode} ${task.name}`);
+  } catch (e: any) {
```
tristan957, line 201: Any better type for this?

src/tests.ts
```diff
@@ -45,13 +46,15 @@ export async function testRunHandler(controller: vscode.TestController, request:
       let duration = Date.now() - starttime;
       run.passed(test, duration);
     } catch (e) {
```
tristan957, line 48: Why not cast e here?

src/tools/muon.ts
```diff
@@ -87,24 +87,22 @@ export async function format(
...
-
-  const undef_muon = { path: undefined, version: undefined };
-
   let stdout: string, stderr: string;
 
   try {
-    ({ stdout, stderr } = await exec(muon_path, ["version"]))
-  } catch (exception) {
-    const { error, stdout, stderr }: { error: cp.ExecException, stdout: string, stderr: string } = exception;
+    ({ stdout, stderr } = await exec(muon_path, ["version"]));
```
tristan957, line 95: Why is the catch on its own line?

src/tools/muon.ts
```diff
@@ -87,24 +87,22 @@ export async function format(
...
+  }
+  catch (e) {
+    const { error, stdout, stderr } = e as ExecResult;
```
tristan957, line 98: Same thing with the cast

src/tasks.ts
lukester1975: Sadly not, but that's TS exceptions for you! Can't specify a type other than any (the default with strict `unknown`  due to `useUnknownInCatchVariables`).

So this just introspects `e` since it seemed pretty unspecified as to what exception type might be caught here (if any...). Plenty of other approaches for somewhat type-safe exception handling, but it's no worse than it was, hopefully, so didn't seem worth going crazy.

src/tests.ts
lukester1975: Same reason. Just asserts it later as `ExecResult` since that's known to be the exception type here from the `exec`.

src/tools/muon.ts
lukester1975: No reason, other than no hook is enforcing style 😁

src/tools/muon.ts
lukester1975: Yes, same!

PR #149, opened 2023-07-11
Add meson.options file type

xclaesse:
-    Add meson.options file type

-    tasks.json: Add missing version field

-    Reformat all json files
    
    Some were using tab indentation, but not all. Use the default json
    formatter from vscode, simply saving the file is reformatting them.


tristan957:
Also add a changelog bullet point for the PR

xclaesse:
> Wanna add a git blame ignore revs file for your reformat commit?

done, please merge without rebase (should be on top of main branch atm) otherwise commit won't match.

> Also add a changelog bullet point for the meson.options commit

done

PR #154, opened 2023-09-27
List tests on activation

ylatuya:
The tests list in the tests controller is only updated when a change is detected. It's now done on activation  to list all tests from the beginning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## next
 
 - Add syntax highlight for `meson.options` files.
+- Fix listing tests in the Test Explorer on activation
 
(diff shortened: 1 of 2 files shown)


ylatuya:
cc: @xclaesse

tristan957:
Could you explain the bug this fixes a little more clearly? I don't quite understand

ylatuya:
In this case, an image is worth a thousand words :)

Before
Screenshot 2023-09-28 at 11 09 15

After
Screenshot 2023-09-28 at 11 12 56

tristan957:
Perfect. Thanks for your work!

PR #162, opened 2023-10-20
Add support for CodeLLDB extension

wolfpld:
Currently, the vscode-meson extension is hardcoded to support only Microsoft's C/C++ extension, which provides the `cppdbg` debugger. However, using the Microsoft extension may not be preferable. An alternative extension, CodeLLDB (https://github.com/vadimcn/codelldb), also provides C/C++ support, including the `lldb` debugger.

These changes add a second debug configuration provider that works with the `lldb` debugger.

These changes also introduce an extension availability check when trying to run tests with debugging. Both the `cppdbg` and `lldb` debuggers are supported, in that order. In addition, a proper error message is now displayed instead of the cryptic "Configured debug type 'cppdbg' is not supported" when no debuggers are available.

Note: The `lldb` debug configuration sets the working directory to `path.dirname(this.path)`, which translates to the parent directory of the build root. This matches the `cppdbg` debug configuration. However, the tests debug configuration falls back to `workspaceRelative(extensionConfiguration("buildFolder")`. Shouldn't this be the same?

### DIFF
--- a/src/configprovider-cppdbg.ts
+++ b/src/configprovider-cppdbg.ts
@@ -2,11 +2,6 @@ import * as vscode from "vscode";
 import { Target } from "./types";
 import { MesonDebugConfigurationProvider } from "./configprovider";
 
-export enum MIModes {
-  lldb = "lldb",
-  gdb = "gdb",
-}
-
 export class DebugConfigurationProviderCppdbg extends MesonDebugConfigurationProvider {
   constructor(path: string) {
     super(path);
@@ -30,7 +25,7 @@ export class DebugConfigurationProviderCppdbg extends MesonDebugConfigurationPro
 
   async createGDBDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
     let debugConfig = await super.createDebugConfiguration(target);
-    debugConfig["MIMode"] = MIModes.gdb;
+    debugConfig["MIMode"] = "gdb";
     debugConfig["setupCommands"] = [
       {
         description: "Enable pretty-printing for gdb",
@@ -43,7 +38,7 @@ export class DebugConfigurationProviderCppdbg extends MesonDebugConfigurationPro
 
   async createLLDBDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
     let debugConfig = await super.createDebugConfiguration(target);
-    debugConfig["MIMode"] = MIModes.lldb;
+    debugConfig["MIMode"] = "lldb";
     return debugConfig;
   }
 
--- a/src/configprovider-cppdbg.ts
+++ b/src/configprovider-cppdbg.ts
@@ -10,6 +10,8 @@ export enum MIModes {
 }
 
 export class DebugConfigurationProviderCppdbg implements vscode.DebugConfigurationProvider {
+  static readonly type = "cppdbg";
+
   private path: string;
 
   constructor(path: string) {
@@ -19,7 +21,7 @@ export class DebugConfigurationProviderCppdbg implements vscode.DebugConfigurati
   async createBaseDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
     const targetName = await getTargetName(target);
     return {
-      type: "cppdbg",
+      type: DebugConfigurationProviderCppdbg.type,
       name: `Debug ${target.name} (cppdbg)`,
       request: "launch",
       cwd: path.dirname(this.path),
--- a/src/configprovider-cppdbg.ts
+++ b/src/configprovider-cppdbg.ts
@@ -1,38 +1,35 @@
 import * as vscode from "vscode";
-import * as path from "path";
-import { getMesonTargets } from "./introspection";
 import { Target } from "./types";
-import { extensionConfiguration, getTargetName } from "./utils";
+import { MesonDebugConfigurationProvider } from "./configprovider";
 
 export enum MIModes {
   lldb = "lldb",
   gdb = "gdb",
 }
 
-export class DebugConfigurationProviderCppdbg implements vscode.DebugConfigurationProvider {
-  static readonly type = "cppdbg";
-
-  private path: string;
-
+export class DebugConfigurationProviderCppdbg extends MesonDebugConfigurationProvider {
   constructor(path: string) {
-    this.path = path;
+    super(path);
   }
 
-  async createBaseDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
-    const targetName = await getTargetName(target);
-    return {
-      type: DebugConfigurationProviderCppdbg.type,
-      name: `Debug ${target.name} (cppdbg)`,
-      request: "launch",
-      cwd: path.dirname(this.path),
-      program: target.filename[0],
-      args: [],
-      preLaunchTask: `Meson: Build ${targetName}`,
-    };
+  override getName(): string {
+    return "cppdbg";
+  }
+
+  override async createDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
+    let debugConfiguration = null;
+    if (target.target_sources?.some((source) => ["cl"].includes(source.compiler[0]))) {
+      debugConfiguration = await this.createMSVCDebugConfiguration(target);
+    } else if (target.target_sources?.some((source) => ["cc", "clang"].includes(source.compiler[0]))) {
+      debugConfiguration = await this.createLLDBDebugConfiguration(target);
+    } else {
+      debugConfiguration = await this.createGDBDebugConfiguration(target);
+    }
+    return debugConfiguration;
   }
 
   async createGDBDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
-    let debugConfig = await this.createBaseDebugConfiguration(target);
+    let debugConfig = await super.createDebugConfiguration(target);
     debugConfig["MIMode"] = MIModes.gdb;
     debugConfig["setupCommands"] = [
       {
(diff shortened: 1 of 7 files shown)


src/configprovider-lldb.ts
```diff
@@ -0,0 +1,65 @@
...
+    let ret: vscode.DebugConfiguration[] = [];
+
+    for (const target of executables) {
+      if (!target.target_sources?.some((source) => ["cpp", "c"].includes(source.language))) {
+        continue;
+      }
+
+      let debugConfiguration = await this.createDebugConfiguration(target);
+      ret.push({ ...configDebugOptions, ...debugConfiguration });
+    }
```
tristan957, line 45: Looks like some of these lets can be const to me.

src/tests.ts
```diff
@@ -99,6 +99,24 @@ export async function testDebugHandler(
...
+    debugType = "cppdbg";
+  } else {
+    const codelldb = vscode.extensions.getExtension("vadimcn.vscode-lldb");
+    if (codelldb && codelldb.isActive) {
+      debugType = "lldb";
+    }
+  }
+
+  if (!debugType) {
+    vscode.window.showErrorMessage("No debugger extension found. Please install one and try again");
```
tristan957, line 115: Can you add a period at the end of this message?

tristan957:
Can you rephrase the first commit message? Something like `Be more specific about the current debug provider" or something would work for me.

tristan957:
Also add a changelog entry.

src/extension.ts
```diff
@@ -41,7 +42,15 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
       "cppdbg",
-      new DebugConfigurationProvider(buildDir),
+      new DebugConfigurationProviderCppdbg(buildDir),
+      vscode.DebugConfigurationProviderTriggerKind.Dynamic,
+    ),
+  );
+
+  ctx.subscriptions.push(
+    vscode.debug.registerDebugConfigurationProvider(
+      "lldb",
```
tristan957, line 52: Can you expose these strings as static public constants on the DebugConfigurationProvider implementations?

src/configprovider-lldb.ts
```diff
@@ -0,0 +1,65 @@
+import * as vscode from "vscode";
+import * as path from "path";
+import { getMesonTargets } from "./introspection";
+import { Target } from "./types";
+import { extensionConfiguration, getTargetName } from "./utils";
+
+export class DebugConfigurationProviderLldb implements vscode.DebugConfigurationProvider {
```
tristan957, line 7: How much of this class is now a copy-paste of the Cppdbg one? Anyway that we could factor the copy-paste into a common abstract base class like "MesonDebugConfigurationProvider"?

tristan957 (commented):
I'll spin a release as soon as I get the requested changes. Thanks for this!

src/tests.ts
wolfpld: There is no period at the end in other error messages in this file. In other places in the code, there's no consistency in how this is handled. I can do this, but fixing it seems to be outside the scope of this PR.

https://github.com/mesonbuild/vscode-meson/blob/2d91eebf3beebfd9429e098d8f67cac1ab88ea7d/src/tests.ts#L54
https://github.com/mesonbuild/vscode-meson/blob/2d91eebf3beebfd9429e098d8f67cac1ab88ea7d/src/tests.ts#L97

tristan957: Yeah keep it as is then. Thanks for checking.

src/extension.ts
wolfpld: Should be fixed by f17751a.

src/configprovider-cppdbg.ts
```diff
@@ -10,6 +10,8 @@ export enum MIModes {
 }
 
 export class DebugConfigurationProviderCppdbg implements vscode.DebugConfigurationProvider {
+  static readonly type = "cppdbg";
```
tristan957, line 13: Rename it to `name`. Thanks for this commit.

wolfpld: Static property 'name' conflicts with built-in property 'Function.name' of constructor function 'DebugConfigurationProviderCppdbg'.

tristan957: type is fine then!

src/configprovider-lldb.ts
wolfpld: Quite a lot apparently. e817bb1 should fix things.

src/configprovider-lldb.ts
wolfpld: This was in the code previously. I fixed it in e817bb1.

src/configprovider-cppdbg.ts
```diff
@@ -0,0 +1,55 @@
+import * as vscode from "vscode";
+import { Target } from "./types";
+import { MesonDebugConfigurationProvider } from "./configprovider";
+
+export enum MIModes {
```
tristan957, line 5: Does this need to be exported anymore?

tristan957:
I think you are missing the MesonDebugConfigurationProvider.

src/configprovider-cppdbg.ts
wolfpld: I don't know. It was exported before and I don't know the language.

src/configprovider.ts
```diff
@@ -4,23 +4,22 @@ import { getMesonTargets } from "./introspection";
...
+  abstract createDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration>;
+
   async createBaseDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
     const targetName = await getTargetName(target);
+    const name = this.getName();
     return {
-      type: "cppdbg",
-      name: `Debug ${target.name}`,
+      type: name,
+      name: `Debug ${target.name} (` + name + `)`,
```
tristan957, line 22: Use `${name}`

src/configprovider.ts
```diff
@@ -70,15 +44,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
...
-      let debugConfiguration = null;
-      if (target.target_sources.some((source) => ["cl"].includes(source.compiler[0]))) {
-        debugConfiguration = await this.createMSVCDebugConfiguration(target);
-      } else if (target.target_sources.some((source) => ["cc", "clang"].includes(source.compiler[0]))) {
-        debugConfiguration = await this.createLLDBDebugConfiguration(target);
-      } else {
-        debugConfiguration = await this.createGDBDebugConfiguration(target);
-      }
-
+      let debugConfiguration = await this.createDebugConfiguration(target);
```
tristan957, line 47: This can also be const right?

src/extension.ts
```diff
@@ -38,13 +39,18 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
 
-  ctx.subscriptions.push(
-    vscode.debug.registerDebugConfigurationProvider(
-      "cppdbg",
-      new DebugConfigurationProvider(buildDir),
-      vscode.DebugConfigurationProviderTriggerKind.Dynamic,
-    ),
-  );
+  const providers = [DebugConfigurationProviderCppdbg, DebugConfigurationProviderLldb];
+
```
tristan957, line 43: Remove this newline

src/configprovider.ts
```diff
@@ -4,23 +4,22 @@ import { getMesonTargets } from "./introspection";
...
+  private readonly path: string;
 
   constructor(path: string) {
     this.path = path;
   }
 
+  abstract getName(): string;
+  abstract createDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration>;
+
   async createBaseDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
```
tristan957, line 17: I would change this function name to `createDebugConfiguration`. Make it not abstract, and use super in the Cppdbg class to access it after overriding it. Does that make sense?

src/configprovider-cppdbg.ts
wolfpld: The documentation for this is at https://code.visualstudio.com/docs/cpp/launch-json-reference#_mimode and it's just a string that can have two values set.

TBH, I see zero value in having this export here. It just makes you do more reading and writing, so it's actually a net negative. I will remove it and just set the raw strings where appropriate.

```ts
export enum MIModes {
  lldb = "lldb",
  gdb = "gdb",
}
```

wolfpld:
I have addressed all of your suggestions.

src/configprovider-cppdbg.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  async createLLDBDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
+    let debugConfig = await super.createDebugConfiguration(target);
+    debugConfig["MIMode"] = "lldb";
+    return debugConfig;
+  }
+
+  async createMSVCDebugConfiguration(target: Target): Promise<vscode.DebugConfiguration> {
+    let debugConfig = await super.createDebugConfiguration(target);
+    debugConfig.type = "cppvsdbg";
```
wolfpld, line 47: This line goes against the whole design, but it was here before. I wonder if this even works, considering that you have to specify each of the debugger types to be handled by the configuration provider, in `package.json`.

tristan957: Would you mind submitting another PR to refactor this after we get this PR in?

wolfpld: I'm not interested in supporting Microsoft's extension, sorry.

tristan957:
I will spin up a release today. Appreciate your work on this.

PR #163, opened 2023-10-23
Add auto-updater for Swift-MesonLSP

JCWasmx86:
This is using a simpler thing for auto-updates. It couples the Swift-MesonLSP version to the vscode-meson version.  (Somewhat based on https://github.com/mesonbuild/vscode-meson/issues/119#issuecomment-1560059245 and the previous comments). This makes sense as it seems like you do a lot of smaller releases.

For testing, you can edit the file: `~/.config/VSCodium/User/globalStorage/mesonbuild.mesonbuild/lsp/version` to e.g. 1.2.3 and then run VSCode. It should automatically update to 2.4..4.

### DIFF
--- a/src/lsp/common.ts
+++ b/src/lsp/common.ts
@@ -4,20 +4,20 @@ import { LanguageServer } from "../types";
 import { SwiftMesonLspLanguageClient } from "./swift-mesonlsp";
 import { Uri } from "vscode";
 
+export function serverToClass(server: LanguageServer): any {
+  switch (server) {
+    case "Swift-MesonLSP":
+      return SwiftMesonLspLanguageClient;
+    default:
+      return null;
+  }
+}
+
 export async function createLanguageServerClient(
   server: LanguageServer,
   download: boolean,
   context: vscode.ExtensionContext,
 ): Promise<LanguageServerClient | null> {
-  const serverToClass = (server: LanguageServer) => {
-    switch (server) {
-      case "Swift-MesonLSP":
-        return SwiftMesonLspLanguageClient;
-      default:
-        return null;
-    }
-  };
-
   const klass = serverToClass(server);
   if (klass == null) {
     return null;
@@ -54,5 +54,5 @@ export async function createLanguageServerClient(
     }
   }
 
-  return new klass(languageServerPath, context);
+  return new klass(languageServerPath, context, klass.version);
 }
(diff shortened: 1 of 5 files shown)


src/lsp/common.ts
```diff
@@ -4,20 +4,20 @@ import { LanguageServer } from "../types";
 import { SwiftMesonLspLanguageClient } from "./swift-mesonlsp";
 import { Uri } from "vscode";
 
+export function serverToClass(server: LanguageServer): any {
```
tristan957, line 7: Does it complain if you make the return value `typeof LanguageServerClient`?

src/lsp/swift-mesonlsp.ts
```diff
@@ -3,6 +3,7 @@ import * as vscode from "vscode";
 
 import { LanguageServerClient } from "../lsp";
 import { Executable } from "vscode-languageclient/node";
+import { getOutputChannel } from "../utils";
```
tristan957, line 6: This import seems superfluous

src/lsp/index.ts
```diff
@@ -193,4 +201,17 @@ export abstract class LanguageServerClient {
     );
     this.ls.start();
   }
+
+  async update(context: vscode.ExtensionContext): Promise<void> {
+    const lspDir = storage.uri(storage.Location.LSP, context).fsPath;
+    const versionFile = path.join(lspDir, "version");
+    if (!fs.existsSync(versionFile)) return; // Either we use binaries from PATH or something is broken.
```
tristan957, line 208: Can you add some whitespace in this function so it isn't a giant block of code

src/lsp/index.ts
```diff
@@ -33,10 +35,16 @@ export abstract class LanguageServerClient {
   protected abstract get debugExe(): Executable;
   protected abstract get runExe(): Executable;
 
-  protected constructor(server: LanguageServer, languageServerPath: vscode.Uri, context: vscode.ExtensionContext) {
+  protected constructor(
+    server: LanguageServer,
+    languageServerPath: vscode.Uri,
+    context: vscode.ExtensionContext,
+    version: string,
```
tristan957, line 42: Let's call this referenceVersion

src/lsp/common.ts
tristan957: Yeah I feel like I tried to fix this before we merged the PR, and I couldn't figure it out then either.

tristan957 (approved):

PR #164, opened 2023-10-23
Send configuration fo mesonbuild to language server, if it changed

JCWasmx86:
Configuration format is like this in your .settings.json
```
{
    "mesonbuild.Swift-MesonLSP": {
        "others": {
            "ignoreDiagnosticsFromSubprojects": true,
        },
        "linting": {
            "disableNameLinting": true,
            "disableCompilerIdLinting": true,
        }
    }
}
```

Testing:
- Remove downloaded language server from the vscode extension
- Use binaries from HEAD of https://github.com/JCWasmx86/Swift-MesonLSP and drop them in PATH
- Use e.g. https://github.com/hse-project/hse

Your diagnostics will look like this:
image

Merge the config listed above with your .settings.json.

Only this error will be left:
image

(A bug on my side, as my heuristics will never be able to guess the sophisticated subdir calculation for hse-python, but that's offtopic here)

Only thing missing here is the auto-completion/diagnostics in the settings.json file.

### DIFF
--- a/src/lsp/index.ts
+++ b/src/lsp/index.ts
@@ -7,6 +7,8 @@ import * as os from "os";
 import * as path from "path";
 import * as vscode from "vscode";
 import {
+  DidChangeConfigurationNotification,
+  DidChangeConfigurationParams,
   Executable,
   LanguageClient,
   LanguageClientOptions,
@@ -15,6 +17,7 @@ import {
 } from "vscode-languageclient/node";
 import * as storage from "../storage";
 import { LanguageServer } from "../types";
+import { getOutputChannel } from "../utils";
 
 export abstract class LanguageServerClient {
   private static readonly clientOptions: LanguageClientOptions = {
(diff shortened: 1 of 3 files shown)


src/lsp/index.ts
```diff
@@ -183,7 +186,8 @@ export abstract class LanguageServerClient {
       debug: this.debugExe,
       transport: TransportKind.stdio,
     };
-
+    let options = LanguageServerClient.clientOptions;
```
tristan957, line 189: I am pretty sure that this can be const

src/lsp/index.ts
```diff
@@ -193,4 +197,12 @@ export abstract class LanguageServerClient {
     );
     this.ls.start();
   }
+
+  async reloadConfig(): Promise<void> {
+    const config = vscode.workspace.getConfiguration(`mesonbuild.${this.server}`);
+    const params: DidChangeConfigurationParams = {
+      settings: config,
+    };
+    await this.ls?.sendNotification(DidChangeConfigurationNotification.type, params);
```
tristan957, line 206: Can this be an `!` instead of a `?`? Can we ever be here if ls is null?

tristan957 (approved):

tristan957:
Let's just add a changelog entry.

tristan957:
It is funny that you are testing with HSE. Fun project to work on. Too bad Micron laid us all off :(

tristan957:
That is fine with me.

PR #184, opened 2023-11-14
Fix "s.sources" is not iterable

tristan957:
When exploring the project tree, I ran into this error.


tristan957:
@xclaesse you mind taking a look at the type changes?

xclaesse:
Sure, LGTM

tristan957:
But does this match what Meson outputs in introspection?

xclaesse:
I think in theory the introspection json should have empty list `"generated_sources" : []`, but I assume since you're doing this PR you found a case it's null?

tristan957:
I only saw the issue with `sources`. I was just being safe with the `generated_sources`.

PR #188, opened 2023-11-15
Sort meson.build selection by directory depth first.

wolfpld:
obraz

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -19,7 +19,7 @@ import { activateLinters } from "./linters";
 import { activateFormatters } from "./formatters";
 import { SettingsKey, TaskQuickPickItem } from "./types";
 import { createLanguageServerClient } from "./lsp/common";
-import { dirname, relative } from "path";
+import { dirname, relative, sep } from "path";
 
 export let extensionPath: string;
 export let workspaceState: vscode.Memento;
@@ -237,8 +237,17 @@ export async function activate(ctx: vscode.ExtensionContext) {
     if (configureOnOpen) {
       let cancel = false;
       if (!configurationChosen && mesonFiles.length > 1) {
-        const items = mesonFiles.map((file, index) => ({ index: index, label: relative(root, file.fsPath) }));
-        items.sort((a, b) => a.label.localeCompare(b.label));
+        const items = mesonFiles.map((file, index) => {
+          const label = relative(root, file.fsPath);
+          return { index, label, depth: label.split(sep).length };
+        });
+        items.sort(function (a, b) {
+          if (a.depth === b.depth) {
+            return a.label.localeCompare(b.label);
+          } else {
+            return a.depth - b.depth;
+          }
+        });
         const selection = await vscode.window.showQuickPick(items, {
           canPickMany: false,
           title: "Select configuration to use.",


src/extension.ts
```diff
@@ -238,7 +238,15 @@ export async function activate(ctx: vscode.ExtensionContext) {
       let cancel = false;
       if (!configurationChosen && mesonFiles.length > 1) {
         const items = mesonFiles.map((file, index) => ({ index: index, label: relative(root, file.fsPath) }));
-        items.sort((a, b) => a.label.localeCompare(b.label));
+        items.sort(function (a, b) {
+          const aDepth = a.label.split("/").length;
+          const bDepth = b.label.split("/").length;
```
tristan957, line 243: Does this work on Windows?

wolfpld: Now it does.

src/extension.ts
```diff
@@ -237,8 +237,19 @@ export async function activate(ctx: vscode.ExtensionContext) {
     if (configureOnOpen) {
       let cancel = false;
       if (!configurationChosen && mesonFiles.length > 1) {
-        const items = mesonFiles.map((file, index) => ({ index: index, label: relative(root, file.fsPath) }));
-        items.sort((a, b) => a.label.localeCompare(b.label));
+        const itemsBase = mesonFiles.map((file, index) => ({ index: index, label: relative(root, file.fsPath) }));
+        let items = itemsBase.map((item) => ({ ...item, depth: item.label.split("/").length }));
```
tristan957, line 241: Shouldn't you use path.sep?

wolfpld: First you need to know it exists. Done.

src/extension.ts
```diff
@@ -237,8 +237,15 @@ export async function activate(ctx: vscode.ExtensionContext) {
     if (configureOnOpen) {
       let cancel = false;
       if (!configurationChosen && mesonFiles.length > 1) {
-        const items = mesonFiles.map((file, index) => ({ index: index, label: relative(root, file.fsPath) }));
-        items.sort((a, b) => a.label.localeCompare(b.label));
+        const itemsBase = mesonFiles.map((file, index) => ({ index: index, label: relative(root, file.fsPath) }));
+        const items = itemsBase.map((item) => ({ ...item, depth: item.label.split(sep).length }));
```
tristan957, line 241: Can this be done more concisely? I don't understand why it is O(2n) instead of O(n)

wolfpld: I don't know a way to access `label` without recomputing `relative` in the first `map` call.

tristan957:
```typescript
const items = mesonFiles.map((file, index) => {
  const label = relative(root, file.fsPath);
  const depth = label.split(sep).length;

  return { index, label, depth };
});
```

wolfpld: Thanks, done.

tristan957:
Can you confirm that this still works with all the changes?

wolfpld:
I always check that things are working before pushing changes. Checking with Windows is a bit more involved as I have to manually build and install the vsix, so the last time I checked was after the `path.sep` changes. Subsequent changes in logic have not affected the way this particular piece of functionality works, so there was no need to check it again.

xclaesse:
The problem is much worse than this. By default it picks a random meson.build file and decide the buildDir must be there: https://github.com/mesonbuild/vscode-meson/blob/main/src/extension.ts#L57. 

If there is a meson.build file at the root, it should use that one a stop looking anywhere else. That's by far the much more common case. If none is found at the root, we could eventually lookup in subdirs and stop recursing into sub-subdir as soon as a meson.build is found.

For example, if we have
- proj1/meson.build
- proj1/foo/meson.build
- proj2/meson.build

Only `proj1/meson.build` and `proj2/meson.build` should be shown. Not `proj1/foo/meson.build`.

xclaesse:
To be clear, if this cannot be solved immediately, we should absolutely revert https://github.com/mesonbuild/vscode-meson/pull/179, or somehow rollback latest extension releases. Current release is completely unusable for everyone.

xclaesse:
Sorry if I sounded harsh, let's just try to fix this quickly :)

wolfpld:
> By default it picks a random meson.build file and decide the buildDir must be there

This is not true.

> If there is a meson.build file at the root, it should use that one a stop looking anywhere else. 

Why?

xclaesse:
> > If there is a meson.build file at the root, it should use that one a stop looking anywhere else.
> 
> Why?

Because when you have a meson.build file at the root, why should I chose another meson.build file? Projects can have hundreds of meson files but only the one at the root matters. 

> > By default it picks a random meson.build file and decide the buildDir must be there
> 
> This is not true.

On first run, it picks `mesonFiles[0]` which can be anything.
```
  const mesonFile = savedMesonFile ?? mesonFiles[0].fsPath;
  const sourceDir = dirname(mesonFile);
  const buildDir = relativeBuildDir(mesonFile);
```

That means that buildDir is not found at the root and thus it wrongly opens the dialog "Meson project detected in this workspace but does not seems to be configured. Would you like VS Code to configure it?". Note that if you click No it would still ask to chose a meson.build file in a completely bogus list where none of them can be used as a project:

image

xclaesse:
To be honest, I'm really not convinced by this feature to begin with. It's really niche to not have the project starting at the root of the repository, but even if it does, why now just opening the subdir in vscode to make it the root of the workspace?

If we really want to support multiple projects in the workspace, do we even have a way to switch between projects? It seems that it save the one you picked on the first run, then you cannot change?

tristan957:
> To be honest, I'm really not convinced by this feature to begin with. It's really niche to not have the project starting at the root of the repository, but even if it does, why not just opening the subdir in vscode to make it the root of the workspace?

I think this is a stopgap for a bit while we don't support multiple projects.

> If we really want to support multiple projects in the workspace, do we even have a way to switch between projects? It seems that it save the one you picked on the first run, then you cannot change?

We can add a switch project command.

tristan957:
I believe I fixed this in https://github.com/mesonbuild/vscode-meson/pull/190.

wolfpld:
> Because when you have a meson.build file at the root, why should I chose another meson.build file?

There are certain configurations where such a choice makes sense. I describe one of them below.

> Projects can have hundreds of meson files but only the one at the root matters.

Sure, but this is the `subdir` case, isn't it? I have stated that I intend to fix this by filtering out all `meson.build` files referenced in this way. It was a bit unfortunate that #179 was merged before I could do this.

> That means that buildDir is not found at the root and thus it wrongly opens the dialog "Meson project detected in this workspace but does not seems to be configured. Would you like VS Code to configure it?".

This would only be a problem on the first run, wouldn't it? The selected `meson.build` file will be saved in a `mesonbuild.mesonFile` memento, so on any further opening of the workspace you will have the correct `buildDir` selected, and you will no longer get the selection prompt.

> Note that if you click No it would still ask to chose a meson.build file in a completely bogus list where none of them can be used as a project

This should not happen, and I have checked that it does not. I think it may be happening now because of this requested change: https://github.com/mesonbuild/vscode-meson/pull/179#pullrequestreview-1730250586

Note that with this PR you will have the root level `meson.build` as the first entry in the list, so you can just hit enter and have the previous behavior.

> It's really niche to not have the project starting at the root of the repository, but even if it does, why not just opening the subdir in vscode to make it the root of the workspace?

This is one of my use cases:

> For example, I need support for selecting a meson.build file that is appropriate for the utility I want to work with. The "one build file to rule them all" approach is a bit naive and not sufficient for what Tracy is doing. For example, the GUI server compiles a source file with a flag to enable calculation of statistics that can be presented to the user, while the CLI capture utility does not need to do these calculations. This results in two separate #ifdefed code paths, and you need to have an appropriate set of defines in compile_commands.json depending on which utility you are working on. Switching between two meson.builds is the only way I can see to do this.

While it is possible to create a root-level meson.build that will compile the same set of sources twice, with different flags, I don't see a way to provide a `compilation_commands.json` file for VS Code that would handle both cases at the same time. Hence the need for a switch.

> If we really want to support multiple projects in the workspace, do we even have a way to switch between projects? It seems that it save the one you picked on the first run, then you cannot change?

At this moment, you would have to delete the build dir to get the choice dialog to reappear. Adding support for switching between projects is on my to-do list, as mentioned in https://github.com/mesonbuild/vscode-meson/pull/179#issuecomment-1810503805.

Currently, I'm a bit blocked by this PR being unmerged, as it fixes an immediate problem resulting from the hasty merge of #179. I don't want to add new features to this PR, and working on two branches at the same time, where one depends on the other, is asking for bugs.

tristan957:
> It was a bit unfortunate that https://github.com/mesonbuild/vscode-meson/pull/179 was merged before I could do this.

In no way did you indicate that to me.

tristan957:
Next time please open a draft. That helps me understand what you mean.

wolfpld:
> In no way did you indicate that to me.

Sorry, I thought this was clear after this question: https://github.com/mesonbuild/vscode-meson/pull/179#issuecomment-1810693725

tristan957:
Yeah, ok. That's my bad. Should I just revert it then?

wolfpld:
I don't see an immediate need for a revert. This PR should fix the main problem we currently have, which is making it easy for people to select the root-level `meson.build`. The filtering and switching features can be added later, as they are not critical to have right now if all you want to do is work with the root-level file.

tristan957:
I think my PR is the better solution, so I am more likely to merge that. We shouldn't care about every Meson file ever, other than for the purposes of file watching.

wolfpld:
Can you explain what your PR does? I believe it filters out all `meson.build` files that are in subdirectories of a directory that already contains a `meson.build` file?

I have a use case where this may not be the optimal solution.

Say I have a project with a root level `meson.build` file. I want to build my application with it and everything is fine here.

My application also depends on an external library. The library lives in a submodule and provides its own `meson.build` file (or files) to build the utilities needed to use it properly. Some source files of this library are shared between the application and the utilities provided by the library, for example to provide a common network ABI.

Now, as I develop some features in this library, I want to be able to quickly switch between the context of the application and the library utilities. For example, I can add some functionality to the library on the application side, and then keep my files open when I switch to working on the library itself.

Now the `meson.build` for the library is no longer selectable, since there is already a root-level `meson.build` for the application, because it is contained in a submodule directory within the workspace.

I would eventually like to modify the `rootMesonFiles()` functionality to work by filtering out the `subdir()` references, but the current solution will do for now.

> I'll keep this open for now, but we may want to just close it.

I think this should be merged in, regardless of your change. In the context of the screenshot I show in the first post in this PR, I would like to have my options sorted by directory depth.

wolfpld:
@xclaesse Can you give me the URL for the repository you referenced that has a lot of `meson.build' files? A real-world example would make it easier for me to find the potential issues.

wolfpld:
> Every meson projects, really... A big one for example: https://gitlab.freedesktop.org/gstreamer/gstreamer.

Thanks.

> One possibility in that case is to parse the meson.build file and see if it starts with `project()` function.

Makes sense.

> In that case, shouldn't you use vscode multiple workspaces feature? Each workspace will have a meson.build at its root, and with my PR you'll be able to select which one you want.

Good question. I will definitely consider this option. It probably removes the need to consider subdirectory `meson.build` files if there's a `meson.build` in a parent directory.

wolfpld:
No longer applicable.

PR #189, opened 2023-11-16
Fix target source tree view.

lukester1975:
Due to https://github.com/mesonbuild/meson/commit/5eb55075.

I guess `types.ts` should be updated, but this at least gets the view working again.


tristan957:
See my other PR in #184

PR #190, opened 2023-11-16
Fix random file being chosen to create project

tristan957:


src/utils.ts
```diff
@@ -179,3 +179,24 @@ export async function useCompileCommands(buildDir: string) {
 export function checkMesonIsConfigured(buildDir: string) {
   return fs.existsSync(path.join(buildDir, "meson-private", "coredata.dat"));
 }
+
+export async function rootMesonFiles(): Promise<vscode.Uri[]> {
+  const allFiles = (await vscode.workspace.findFiles("**/meson.build")).sort(
+    (a, b) => a.fsPath.length - b.fsPath.length,
```
xclaesse, line 185: That's the path lenght in chars? Won't it make `a/b/meson.build` before `long_dir_name/meson.build`?

xclaesse:
In python I would do:
```python
import os
for root, dirs, files in os.walk("."):
    if "meson.build" in files:
        print(f"Root at {root}")
        # Do not look into subdirs if we found a meson.build file
        dirs.clear()
```

src/utils.ts
tristan957: Hmm you're right. I think I need to sort by file path components

wolfpld: If I understand the code correctly, I don't think this would be a problem. The path-length check should be sufficient, since there's no scenario in which you would have found a parent-level `meson.build' after processing a subdirectory of said parent.

PR #191, opened 2023-11-16
WIP: run all tests button and context menus

lukester1975:
First commit is the run/run all buttons in the meson view.

Second commit adds context menus all over the shop. Makes more sense on my old multiple builddir branch (e.g. I had hacked a very unpleasant debug command, but given the debug provider stuff, doesn't seem worth it unless that can be hooked in to; haven't looked), but maybe worth adding some now.

Personally, I don't like that clicking on a target builds it. Could add an inline button next to the `open meson.build` and just leave clicking on the node to expand it.

Also moves the reconfigure button (if/when multiple root/builddirs are added, it doesn't make sense to have a reconfigure button in the sidebar's header).

Build icon came from the cmake plugin. Whether that would need attribution somewhere I've no idea...

Cheers


package.json
```diff
@@ -87,6 +87,16 @@
       {
         "command": "mesonbuild.restartLanguageServer",
         "title": "Meson: Restart Language Server"
+      },
+      {
+        "command": "mesonbuild.node.runAll",
```
tristan957, line 116: It feels like we should mention tests in the title and the command. Should we add separate commands for benchmarks?

src/treeview/nodes/tests.ts
```diff
@@ -21,39 +26,58 @@ export class TestRootNode extends BaseNode {
...
+    item.contextValue = "meson-test-root";
+
     return item;
   }
 
   override getChildren() {
     return this.tests.map((test) => new TestNode(this.id, test, this.isBenchmark));
   }
+
+  run() {
```
tristan957, line 39: Can you add a return type?

tristan957 (commented):
Mind adding a picture of what this looks like?

package.json
lukester1975: That command is only attached to a test (or benchmark) node as an inline button and context menu entry, so making the title "Run all tests" is a little redundant IMHO. Plus like you say, then requires separate commands for benchmarks.

src/treeview/nodes/tests.ts
lukester1975: Can do, though it's from the `IRunnable` interface; not sure what the gain is other than more stuff to edit if the interface changes.

lukester1975:
run
build

I put back the header configure button as if there's no build dir, there would be no configure button... Was different with the multiple build dir support as there was a project-level node, with build dirs beneath it.

src/treeview/nodes/tests.ts
tristan957: To me, the gain is explicit return types. Helps when reading code. Everyone has different preferences though.

package.json
tristan957: Do you not see value in separating tests and benchmarks?

lukester1975: I haven't merged the nodes together (that would be doubly silly as I separated them a few years ago...). This `runAll` command just executes "on" either the tests node or the benchmarks node.

b

tristan957:
This looks solid to me. Are you ready to go?

lukester1975:
Two things:

1. > Personally, I don't like that clicking on a target builds it. Could add an inline button next to the open meson.build and just leave clicking on the node to expand it.

Feels like nowhere else in the ui does clicking on a non-leaf node not expand it. But maybe there's precedent elsewhere...

2. build icon attribution. I couldn't see anything in the cmake repo.

Cheers

tristan957:
I agree with you on 1. If you want to add a build button, I think that would be a great addition!

lukester1975:
1. OK. Might get to it later.

It was cmake tools.

lukester1975:
OK added a commit with that change.

I need to take a look at the icon though. It's ugly when hovering over a selected target node (OK with dark theme):

ugly

lukester1975:
BTW, icon is ugly in cmake plugin too.

Dark mode is fine.

PR #192, opened 2023-11-16
Run muon format from root of workspace.

lukester1975:
Any reason not to do this? Allows a workspace root relative muon config file path.

Thanks

### DIFF
--- a/src/tools/muon.ts
+++ b/src/tools/muon.ts
@@ -42,7 +42,7 @@ export async function lint(muon: Tool, root: string, document: vscode.TextDocume
   return diagnostics;
 }
 
-export async function format(muon: Tool, document: vscode.TextDocument): Promise<vscode.TextEdit[]> {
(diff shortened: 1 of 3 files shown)


src/tools/muon.ts
```diff
@@ -42,7 +42,7 @@ export async function lint(muon: Tool, root: string, document: vscode.TextDocume
   return diagnostics;
 }
 
-export async function format(muon: Tool, document: vscode.TextDocument): Promise<vscode.TextEdit[]> {
+export async function format(muon: Tool, root: string, document: vscode.TextDocument): Promise<vscode.TextEdit[]> {
```
tristan957, line 45: Wanna call this argument sourceRoot?

lukester1975: Only if `lint()` is changed accordingly...

tristan957 (approved):

PR #195, opened 2023-11-17
Refactor support for multiple Meson projects

xclaesse:
There were many issues, including:
- The most common case is to have a meson.build at the root of the workspace. In that case we should not ask user anything.
- Only the first meson.build in a subdir tree is useful, others are unlikely to be projects.
- We need a way to never ask again to select a meson file, especially important when opening meson project itself which does not have a meson.build at the root, but many projects in test cases.
- If the user cancels selection, the meson extension cannot do anything and should be hidden.
- Selecting a meson file and configuring the project are 2 distinct questions. The user may want to select a file but not let vscode configure it. It is common to configure from an external terminal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -405,6 +405,38 @@
         {
           "command": "mesonbuild.openBuildFile",
           "when": "false"
+        },
+        {
+          "command": "mesonbuild.reconfigure",
+          "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.clean",
+          "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.build",
+          "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.test",
+          "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.run",
+          "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.install",
+          "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.benchmark",
+          "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.restartLanguageServer",
+          "when": "mesonbuild.hasProject"
         }
       ]
     },
--- a/package.json
+++ b/package.json
@@ -87,11 +87,20 @@
       {
         "command": "mesonbuild.restartLanguageServer",
         "title": "Meson: Restart Language Server"
+      },
+      {
+        "command": "mesonbuild.selectRootDir",
+        "title": "Meson: Select Project Root Directory"
       }
     ],
     "configuration": {
       "title": "Meson build configuration",
       "properties": {
+        "mesonbuild.selectRootDir": {
+          "type": "boolean",
+          "default": true,
+          "description": "Ask to select a Meson project root directory when more than one project is detected."
+        },
         "mesonbuild.configureOnOpen": {
           "type": [
             "boolean",
@@ -437,6 +446,10 @@
         {
           "command": "mesonbuild.restartLanguageServer",
           "when": "mesonbuild.hasProject"
+        },
+        {
+          "command": "mesonbuild.selectRootDir",
+          "when": "mesonbuild.hasMultipleProjects"
         }
       ]
     },
--- a/src/dialogs.ts
+++ b/src/dialogs.ts
@@ -1,4 +1,5 @@
 import * as vscode from "vscode";
+import * as path from "path";
 import { extensionConfiguration, extensionConfigurationSet } from "./utils";
 import { SettingsKey } from "./types";
 
@@ -63,4 +64,57 @@ export async function askShouldDownloadLanguageServer(): Promise<boolean> {
   }
 
   return false;
+}
+
+export async function askSelectRootDir(): Promise<boolean> {
+  const selectRootDir = extensionConfiguration(SettingsKey.selectRootDir);
+
+  if (!selectRootDir) return false;
+
+  enum Options {
+    yes = "Yes",
+    no = "No",
+    never = "Never",
+  }
+
+  const response = await vscode.window.showInformationMessage(
+    "Multiple Meson projects detected, select one?",
+    ...Object.values(Options),
+  );
+
+  switch (response) {
+    case Options.yes:
+      return true;
+    case Options.no:
+      return false;
+    case Options.never:
+      extensionConfigurationSet(SettingsKey.selectRootDir, false, vscode.ConfigurationTarget.Workspace);
+      return false;
+  }
+
+  return false;
+}
+
+export async function selectRootDir(rootDirs: string[]): Promise<string | undefined> {
+  // TODO: What label to use when there is more than one workspace?
+  const root = vscode.workspace.rootPath ?? "";
+  const items = rootDirs.map((file, index) => ({ index: index, label: path.relative(root, file) }));
+  items.sort((a, b) => {
+    const aComponents = a.label.split(path.sep).length;
+    const bComponents = b.label.split(path.sep).length;
+    if (aComponents == bComponents) return a.label.localeCompare(b.label);
+    return aComponents - bComponents;
+  });
+  const selection = await vscode.window.showQuickPick(items, {
+    canPickMany: false,
+    title: "Select configuration to use.",
+    placeHolder: "path/to/meson.build",
+  });
+  if (selection) return rootDirs[selection.index];
+  return undefined;
+}
+
+export async function askAndSelectRootDir(rootDirs: string[]): Promise<string | undefined> {
+  if (await askSelectRootDir()) return selectRootDir(rootDirs);
+  return undefined;
 }
--- a/src/dialogs.ts
+++ b/src/dialogs.ts
@@ -96,9 +96,13 @@ export async function askSelectRootDir(): Promise<boolean> {
 }
 
 export async function selectRootDir(rootDirs: string[]): Promise<string | undefined> {
-  // TODO: What label to use when there is more than one workspace?
-  const root = vscode.workspace.rootPath ?? "";
-  const items = rootDirs.map((file, index) => ({ index: index, label: path.relative(root, file) }));
+  // Append meson.build to directories otherwise asRelativePath() returns the
+  // absolute path when the path is the root of a workspace. Even if we fix that,
+  // it would leaves us with an empty string which is not better.
+  const items = rootDirs.map((file, index) => ({
+    index: index,
+    label: vscode.workspace.asRelativePath(path.join(file, "meson.build")),
+  }));
   items.sort((a, b) => {
     const aComponents = a.label.split(path.sep).length;
     const bComponents = b.label.split(path.sep).length;
--- a/src/dialogs.ts
+++ b/src/dialogs.ts
@@ -1,0 +1,66 @@
+import * as vscode from "vscode";
+import { extensionConfiguration, extensionConfigurationSet } from "./utils";
+import { SettingsKey } from "./types";
+
+export async function askConfigureOnOpen(): Promise<boolean> {
+  const configureOnOpen = extensionConfiguration(SettingsKey.configureOnOpen);
+
+  if (typeof configureOnOpen === "boolean") return configureOnOpen;
+
+  enum Options {
+    yes = "Yes",
+    always = "Always",
+    no = "No",
+    never = "Never",
+  }
+
+  const response = await vscode.window.showInformationMessage(
+    "Meson project detected in this workspace but does not seems to be configured. Would you like VS Code to configure it?",
+    ...Object.values(Options),
+  );
+
+  switch (response) {
+    case Options.no:
+      return false;
+    case Options.never:
+      extensionConfigurationSet(SettingsKey.configureOnOpen, false, vscode.ConfigurationTarget.Workspace);
+      return false;
+    case Options.yes:
+      return true;
+    case Options.always:
+      extensionConfigurationSet(SettingsKey.configureOnOpen, true, vscode.ConfigurationTarget.Workspace);
+      return true;
+  }
+
+  return false;
+}
+
+export async function askShouldDownloadLanguageServer(): Promise<boolean> {
+  const downloadLanguageServer = extensionConfiguration(SettingsKey.downloadLanguageServer);
+
+  if (typeof downloadLanguageServer === "boolean") return downloadLanguageServer;
+
+  enum Options {
+    yes = "Yes",
+    no = "Not this time",
+    never = "Never",
+  }
+
+  const response = await vscode.window.showInformationMessage(
+    "Should the extension try to download the language server?",
+    ...Object.values(Options),
+  );
+
+  switch (response) {
+    case Options.yes:
+      extensionConfigurationSet(SettingsKey.downloadLanguageServer, true, vscode.ConfigurationTarget.Global);
+      return true;
+    case Options.never:
+      extensionConfigurationSet(SettingsKey.downloadLanguageServer, false, vscode.ConfigurationTarget.Global);
+      return false;
+    case Options.no:
+      return false;
+  }
+
+  return false;
+}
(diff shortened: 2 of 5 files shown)


xclaesse:
CC @tristan957 @wolfpld. Could you please test this PR with your multiple projects case?

tristan957:
I'll probably take a look Saturday or Sunday. Does the current situation with 1.18.1 work for you?

wolfpld:
I think we have some significant communication issues here.

This is the first time you mention having a build dir preconfigured before opening the workspace in VS Code. Only now can I see what your comments about preferring a root-level `meson.build` really meant. To me, the sort order of the initial list was irrelevant, as it was sorted later when it was to be shown to the user. Yes, your use case is important to preserve, and it was not properly handled by my changes, but instead of properly describing what the problem was, you chose to hasten the changes, resulting in a rushed #190, as can be seen by reading your own post-merge comments there. The alternative I presented, which by the way was waiting half a day, would give you the inconvenience of having to choose the default option from a list *once*. Sorry, but I don't see that as a major deal breaker.

Nota bene, if we are talking about breaking workflows, the workflow in which you select to configure the workspace, in the very dialog you are refactoring in this PR, has been broken for half a year, since May, and no one noticed.

Regarding your question, I don't know how these changes affect me. Because of the communication problems I do not know where the development is going. I asked some questions in the PRs yesterday and I am still waiting for your answer. At this point, I don't know if I want to continue putting time and effort into contributing to this repository, or if I should just fork, do what I need for my use case, and rely on that instead.

Ultimately, as @tristan957 said, you "definitely need more contributors to this extension". It's up to you how you want to handle that.

xclaesse:
@wolfpld sorry if it wasn't clear what was broken, it seemed pretty obvious to me since just opening any project results in the mentioned problems.

In any case, this PR fixes tones of things, not all related to multi projects repository, so I think it would be a good improvements anyway.

wolfpld:
> I think it would be a good improvements anyway.

I agree.

> * The most common case is to have a meson.build at the root of the workspace. In that case we should not ask user anything.
> * Only the first meson.build in a subdir tree is useful, others are unlikely to be projects.

I wouldn't say that my use cases support these statements, but at this point it doesn't matter. There are things that need to be ironed out in my implementation and your proposed solution fixes most of the pressing issues (albeit not in an ideal way from my point of view).

> * We need a way to never ask again to select a meson file, especially important when opening meson project itself which does not have a meson.build at the root, but many projects in test cases.
> * If the user cancels selection, the meson extension cannot do anything and should be hidden.

As mentioned earlier, this was the result of a well-intentioned change (`if a === true` => `if a`) where the variable type was not just boolean. This flew under the radar and broke the dialog.

> Selecting a meson file and configuring the project are 2 distinct questions. The user may want to select a file but not let vscode configure it. It is common to configure from an external terminal.

Would that be to set things such as build type, or such? Maybe this should be exposed in the GUI somehow?

wolfpld:
Some relevant notes about the internal implementation of VS Code mementos:

The memento data is stored in `%APPDATA%/Code/User/workspaceStorage/<workspace hash>/state.vscdb`. The `workspace.json` file in the same directory allows to identify which workspace it is.

The `state.vscdb` file is a sqlite database with `ItemTable` as the only table. This table has two columns: `key`, a unique text, and `value`, a blob.

All mementos used by this extension are stored in a row with the key `mesonbuild.mesonbuild`. The `value` column stores a json data with key-value pairs as specified in the code.

```
sqlite> select value from ItemTable where key='mesonbuild.mesonbuild';
{"mesonbuild.buildDir":"/home/wolf/meson-test-repo-multi/a/b/c/d/e/build","mesonbuild.sourceDir":"/home/wolf/meson-test-repo-multi/a/b/c/d/e","mesonbuild.mesonFile":"/home/wolf/meson-test-repo-multi/a/b/c/d/e/meson.build"}
```

The memento state can be reset with the following command: `delete from ItemTable where key='mesonbuild.mesonbuild';`.

tristan957:
You left a merge conflict in

src/extension.ts
```diff
@@ -21,50 +20,58 @@ import { activateFormatters } from "./formatters";
...
+      let newRootDir = await selectRootDir(rootDirs);
+      if (newRootDir != rootDir) {
+        workspaceState.update("mesonbuild.sourceDir", newRootDir);
+        vscode.commands.executeCommand("workbench.action.reloadWindow");
+      }
+    }),
+  );
+
+  getOutputChannel().appendLine(`Meson project root: ${rootDir}`);
+  vscode.commands.executeCommand("setContext", "mesonbuild.hasProject", rootDir != undefined);
```
tristan957, line 65: Use !==

src/utils.ts
```diff
@@ -180,23 +180,28 @@ export function checkMesonIsConfigured(buildDir: string) {
...
-  for (const a of allFiles) {
-    if (rootFiles.length === 0) {
-      rootFiles.push(a);
-      continue;
+export async function mesonRootDirs(): Promise<string[]> {
+  let rootDirs: string[] = [];
+  let pending: vscode.Uri[] = [];
+  vscode.workspace.workspaceFolders?.forEach((i) => pending.push(i.uri));
+  while (true) {
+    let d = pending.pop();
```
tristan957, line 188: Looks like this can be const

src/utils.ts
```diff
@@ -180,23 +180,28 @@ export function checkMesonIsConfigured(buildDir: string) {
...
-
-  const rootFiles: vscode.Uri[] = [];
-  for (const a of allFiles) {
-    if (rootFiles.length === 0) {
-      rootFiles.push(a);
-      continue;
+export async function mesonRootDirs(): Promise<string[]> {
+  let rootDirs: string[] = [];
+  let pending: vscode.Uri[] = [];
+  vscode.workspace.workspaceFolders?.forEach((i) => pending.push(i.uri));
```
tristan957, line 186: Use an `!` instead of `?`

package.json
```diff
@@ -87,11 +87,20 @@
       {
         "command": "mesonbuild.restartLanguageServer",
         "title": "Meson: Restart Language Server"
+      },
+      {
+        "command": "mesonbuild.selectRootDir",
+        "title": "Meson: Select another project root dir"
```
tristan957, line 93: `Meson: Select Root Directory"

package.json
```diff
@@ -87,11 +87,20 @@
...
       }
     ],
     "configuration": {
       "title": "Meson build configuration",
       "properties": {
+        "mesonbuild.selectRootDir": {
+          "type": "boolean",
+          "default": true,
+          "description": "Should Meson ask to select a Meson project root directory when more than one project is detected."
```
tristan957, line 102: Remove the first two words of the description

tristan957:
It would also  be great to get some of this split out into separate commits. Some of the changes seem unrelated to multiple meson projects. I like the refactoring you did with dialogs.ts

src/extension.ts
xclaesse: OOC, I never understood === and !==, what are those operators?

src/utils.ts
xclaesse: OOC, what does those operators does?

src/extension.ts
xclaesse: Nevermind, understood: https://stackoverflow.com/questions/57125700/why-use-triple-equal-in-typescript

xclaesse: done

src/utils.ts
xclaesse: done

package.json
xclaesse: done

package.json
xclaesse: done

xclaesse:
> You left a merge conflict in

Fixed.



> It would also be great to get some of this split out into separate commits.

done.

src/utils.ts
xclaesse: done

src/extension.ts
```diff
@@ -20,51 +19,58 @@ import { activateLinters } from "./linters";
...
   }
 
-  const mesonFile = savedMesonFile ?? mesonFiles[0].fsPath;
-  const sourceDir = dirname(mesonFile);
-  const buildDir = getBuildDirectory(sourceDir);
+  ctx.subscriptions.push(
+    vscode.commands.registerCommand("mesonbuild.selectRootDir", async () => {
+      let newRootDir = await selectRootDir(rootDirs);
+      if (newRootDir != rootDir) {
+        workspaceState.update("mesonbuild.sourceDir", newRootDir);
```
wolfpld, line 57: This won't be saved reliably without `await`, because the `reloadWindow` command executed below will remove the workspace state object immediately.

(@tristan957 Please comment if this makes sense from the language point of view.)

xclaesse: Makes sense, fixed.

wolfpld:
I found a problem that was probably always present in the code, but is now better exposed.

Assuming you have selected a project root, if the "Meson project detected in this workspace but does not seem to be configured..." dialog appears and you click the "no" button or the "x" close button, the project is not configured and the extension continues to work as expected. For example, you can "select project root directory" and it will work.

However, if you close the initial dialog with the Esc key instead, the extension will not respond. No command (e.g. "select root dir" or "reconfigure") will work in this case.

---

Another problem: When the "select project root directory" command is selected (but not when it is displayed in response to the "multiple meson projects detected..." dialog) and you click outside the selection popup, the popup closes and the VS Code window reloads, and the "multiple meson projects detected..." dialog reappears, even if you have already made this selection.

xclaesse:
> However, if you close the initial dialog with the Esc key instead, the extension will not respond. No command (e.g. "select root dir" or "reconfigure") will work in this case.

I think this is how vscode is expected to work... https://github.com/Microsoft/vscode/issues/62853. Pressing escape hide that notification, but it's still running. In our code that means we are still awaiting for the response.

> When the "select project root directory" command is selected (but not when it is displayed in response to the "multiple meson projects detected..." dialog) and you click outside the selection popup, the popup closes and the VS Code window reloads, and the "multiple meson projects detected..." dialog reappears, even if you have already made this selection.

Right, if `selectRootDir()` returns undefined we should consider the user cancelled the request and keep using current sourceDir. But we probably should add a special value in the dropdown menu to select "no meson project". For now I updated this PR to not update when it returns undefined.

src/dialogs.ts
```diff
@@ -0,0 +1,66 @@
+import * as vscode from "vscode";
+import { extensionConfiguration, extensionConfigurationSet } from "./utils";
+import { SettingsKey } from "./types";
+
+export async function askConfigureOnOpen(): Promise<boolean> {
+  let configureOnOpen = extensionConfiguration(SettingsKey.configureOnOpen);
```
tristan957, line 7: const

src/extension.ts
```diff
@@ -202,38 +202,7 @@ export async function activate(ctx: vscode.ExtensionContext) {
...
-          configureOnOpen = true;
-          break;
-
-        case Options.always:
-          extensionConfigurationSet(SettingsKey.configureOnOpen, true, vscode.ConfigurationTarget.Workspace);
-          configureOnOpen = true;
-          break;
-      }
-    }
+    let configureOnOpen = await askConfigureOnOpen();
```
tristan957: Seems like this can be const too?

src/dialogs.ts
```diff
@@ -0,0 +1,66 @@
...
+      return true;
+    case Options.always:
+      extensionConfigurationSet(SettingsKey.configureOnOpen, true, vscode.ConfigurationTarget.Workspace);
+      return true;
+  }
+
+  return false;
+}
+
+export async function askShouldDownload(): Promise<boolean> {
```
tristan957, line 39: s/askShouldDownload/ashShouldDownloadLanugageServer

package.json
```diff
@@ -87,11 +87,20 @@
       {
         "command": "mesonbuild.restartLanguageServer",
         "title": "Meson: Restart Language Server"
+      },
+      {
+        "command": "mesonbuild.selectRootDir",
+        "title": "Meson: Select project root directory"
```
tristan957, line 93: Can you make this command title case similar to the others?

src/dialogs.ts
```diff
@@ -64,3 +65,51 @@ export async function askShouldDownload(): Promise<boolean> {
...
+    case Options.never:
+      extensionConfigurationSet(SettingsKey.selectRootDir, false, vscode.ConfigurationTarget.Workspace);
+      return false;
+  }
+
+  return false;
+}
+
+export async function selectRootDir(rootDirs: string[]): Promise<string | undefined> {
+  // TODO: What label to use when there is more than one workspace?
```
tristan957, line 99: What shoule we do with this TODO?

tristan957 (commented):
Just some nit picks, then let's merge it and tag a release tomorrow.

package.json
tristan957: Should this maybe just be "Select Project"?

src/dialogs.ts
wolfpld: I think this is the use case:

obraz

The `meson-test-build-multi` is a root of the workspace, and its subdirectory `build` is another root, as you can see in the explorer pane on the right. The subdirectory `and/this/is` has also been added as another root. Finally, the external `meson-test-repo-single` directory has been added as yet another root.

So far, the only problem I see here is that the root directory `meson.build` is left unlabeled. Maybe we should use `/` instead?

---

Another issue is that the list of available configurations is only created when the extension is initialized. If you add VS Code provides an `onDidChangeWorkspaceFolders` event, we should probably listen to it and react accordingly.

src/dialogs.ts
xclaesse: done

src/extension.ts
xclaesse: Was already changed in subsequent commits.

src/dialogs.ts
xclaesse: done

src/dialogs.ts
xclaesse: Ok, turns out there is actually a method specifically for our use-case: `vscode.workspace.asRelativePath()`. I also added back the `/meson.build` suffix as in the original code because that fix 2 problems at once: no more empty entries, and asRelativePath() was returning absolute path when the path is a workspace root.

wolfpld: This is the list I get for the above scenario, after the changes. Note that the root `meson.build` is no longer the default on the list.

obraz

xclaesse: > Another issue is that the list of available configurations is only created when the extension is initialized

Indeed, fixed by calling `mesonRootDirs()` again when running the command to get the new list. I don't think we need to be notified for this.

wolfpld: > Indeed, the sorting is broken, but that's handled in https://github.com/mesonbuild/vscode-meson/pull/199, so I think we can leave that out of scope here.

I'm not sure this fix will work properly. It counts the number of directories in a path and sorts by that, but with these new changes, the root `meson.build` no longer has zero directories in its path*. I think it may get mixed up with the rest of the found files.

For example, if I add a new `contrib` workspace root that has `meson.build` at its root, I get this:

obraz

And now both `contrib/meson.build`** and `meson-test-repo-multi/meson.build` are at the same directory depth, and if sorted by name, the contrib one will show up first in the list.

*) The suggested solution of using `/` instead would also break this.
**) Actually `meson-test-repo-multi/contrib/meson.build`.

package.json
tristan957: I feel like "Project" is probably better just because our Memento is `mesonbuild.hasProject`. Plus, I feel like thinking in terms of root directories is a little too low level.

src/dialogs.ts
xclaesse: TBH, I haven't looked much what that other PR does, but we can discuss there implementation details.

xclaesse: Note that sorting should happen before appending `/meson.build` suffix, that part is really just a display label thing. I think that way it will work.

wolfpld: The problem is that `vscode.workspace.asRelativePath()` doesn't seem to return the root level `meson.build` as a null path, instead returning the workspace root name. If roots are to be treated as equivalent, there's not much we can do here, since there is no "primary" root to select.

In the previous approach, the first root was special because all paths were computed relative to it, giving the first root `meson.build` preference.

xclaesse:
Rebased and it uses @tristan957's sorting algo. Good to go?

PR #198, opened 2023-11-22
Look up relative paths to the language server in `PATH` first

paveloom:
In the case the custom path to the language server is relative, look it up in `PATH` first. Useful if the name of the language server is `swift-mesonlsp` instead of `Swift-MesonLSP`.

### DIFF
--- a/src/lsp/index.ts
+++ b/src/lsp/index.ts
@@ -148,8 +148,15 @@ export abstract class LanguageServerClient {
 
   static resolveLanguageServerPath(server: LanguageServer, context: vscode.ExtensionContext): vscode.Uri | null {
     const config = vscode.workspace.getConfiguration("mesonbuild");
-    if (config["languageServerPath"] !== null && config["languageServerPath"] != "")
-      return vscode.Uri.from({ scheme: "file", path: config["languageServerPath"] });
+
+    const configLanguageServerPath = config["languageServerPath"];
+    if (configLanguageServerPath !== null && configLanguageServerPath != "") {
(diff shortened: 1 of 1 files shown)


tristan957:
I guess I am trying to understand a little bit more. Could provide some more context? :)

paveloom:
Currently, if one sets a custom language server path in the extension options, and they use a binary name (e.g., `swift-mesonlsp`), this will not work, as the path will be resolved as `file:///swift-mesonlsp`:

```
[Error - 7:18:36 PM] Meson Language Server (Swift-MesonLSP) client: couldn't create connection to server.
Launching server using command /swift-mesonlsp failed. Error: spawn /swift-mesonlsp ENOENT
```

This is handled here:

https://github.com/mesonbuild/vscode-meson/blob/40d59527a78b017b35f41f3f92451f8e506227ca/src/lsp/index.ts#L151-L152

That's because [`vscode.Uri.from`](https://code.visualstudio.com/api/references/vscode-api#Uri.from) expects already resolved paths.

Now, if one doesn't use a custom server path, then the extension will eventually look up `PATH` for the binary:

https://github.com/mesonbuild/vscode-meson/blob/40d59527a78b017b35f41f3f92451f8e506227ca/src/lsp/index.ts#L157-L158

I copied this logic to solve the above scenario.

paveloom:
Well, I just have it like that in [Nixpkgs](https://github.com/NixOS/nixpkgs/pull/269042). Might change it, though, so as to align with your packages.

tristan957:
I see. Thanks for the context. Looks like this is good to merge. What do you think @JCWasmx86?

PR #199, opened 2023-11-23
Sort root Meson files by number of path components

tristan957:
Previous implementation was way too naive.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -181,9 +181,14 @@ export function checkMesonIsConfigured(buildDir: string) {
 }
 
 export async function rootMesonFiles(): Promise<vscode.Uri[]> {
-  const allFiles = (await vscode.workspace.findFiles("**/meson.build")).sort(
-    (a, b) => a.fsPath.length - b.fsPath.length,
-  );
+  const allFiles = (await vscode.workspace.findFiles("**/meson.build")).sort((a, b) => {
+    const aComponents = a.fsPath.split(path.sep).length;
+    const bComponents = b.fsPath.split(path.sep).length;
+
+    if (aComponents == bComponents) return a.fsPath.localeCompare(b.fsPath);
+
+    return aComponents - bComponents;
+  });
 
   const rootFiles: vscode.Uri[] = [];
   for (const a of allFiles) {


xclaesse:
I think we should instead merge https://github.com/mesonbuild/vscode-meson/pull/195. It also fix that function, and the advantage is it does not need to glob the whole directory, in the most common case you just need to check existence in the root directory.

wolfpld:
Note that I have also privately rebased #188 on top of #195, assuming that this is the way forward to continue the work. Can we agree on how to move this forward?

tristan957:
Let's move forward with #195, but I don't think it is the future of the extension. I think we really just need proper multi-root workspace support.

tristan957:
I am going to submit a bug fix release, and I would like to just merge this since it is a bug fix outside of #195. Is that fine with everybody?

tristan957:
Alternative @xclaesse could split his fix out into a separate PR that I can merge.

wolfpld:
> Is that fine with everybody?

It's basically what I'm also doing, so yes.

src/utils.ts
```diff
@@ -181,9 +181,14 @@ export function checkMesonIsConfigured(buildDir: string) {
...
 
 export async function rootMesonFiles(): Promise<vscode.Uri[]> {
-  const allFiles = (await vscode.workspace.findFiles("**/meson.build")).sort(
-    (a, b) => a.fsPath.length - b.fsPath.length,
-  );
+  const allFiles = (await vscode.workspace.findFiles("**/meson.build")).sort((a, b) => {
+    const aComponents = a.fsPath.split(path.sep).length;
+    const bComponents = b.fsPath.split(path.sep).length;
+
+    if (aComponents == bComponents) return a.fsPath.localeCompare(b.fsPath);
```
xclaesse, line 188: I think it's a bit more complicated than this. What you want is to sort alphabetically each dir part, in pseudo-code:
```
const aComponents = a.fsPath.split(path.sep);
const bComponents = b.fsPath.split(path.sep);
len = min(aComponents.lenght,bComponents.lenght )
for (i=0; i<len; i++) {
  v = strcmp(aComponents[i], bComponents[i])
  if v!=0
    return v;
}
return aComponents.lenght - bComponents.lenght;
```

xclaesse: In python that's what pathlib.Path does.

wolfpld: Can you give an example where the original solution fails? I think sorting by whole path should be equivalent to sorting by each path component?

xclaesse: I believe the code above would sort like that:
```
/z
/a/many/parts
/z/even/more/parts
```

But that should be sorted as
```
/a/many/parts
/z
/z/even/more/parts
```

```python
>>> sorted([Path('/z'), Path('/a/many/parts'), Path('/z/even/more/parts')])
[PosixPath('/a/many/parts'), PosixPath('/z'), PosixPath('/z/even/more/parts')]
```

xclaesse: Hm, actually I take that back, because in our case we can't have both `/z` and `/z/even/more/parts`.

xclaesse: But still I don't think it's logical to have `/b` before `/a` in that case:
```
/b
/a/many/parts
/z/even/more/parts
```

wolfpld: It is intentional this way, to make the root-level `meson.build` the first option, or to promote the project build files against the contrib ones, e.g.

```
library/meson.build
utils/meson.build
contrib/some-lib/meson.build
```

If that's not something we want to consider, the alphabetic sort should give you the same results as sorting by path components.

```
[20:24 wolf@oberon:~]% cat a
/z
/a/many/parts
/z/even/more/parts
[20:25 wolf@oberon:~]% sort a
/a/many/parts
/z
/z/even/more/parts
```

tristan957:
> > Is that fine with everybody?
> 
> It's basically what I'm also doing, so yes.

@wolfpld definitely didn't want to steal your thunder, just didn't think it was the right place for the fix. Opened this PR is a haste to get a bug fix out, but obviously haven't merged it.

src/utils.ts
xclaesse: To be fair, I'm nitpicking, it's really a niche case after all. I guess this is good enough :)

xclaesse: Ok, right, forget me. Sorry for the noise :D

PR #202, opened 2023-11-26
Use shell mode for executing commands

aidenfoxx:
In Windows when running the plugin with the MSYS2 Meson package, the plugin throws the error "Meson version doesn't match expected output:". After some investigation I found that the response from `execFile` was returning: `error: null`, `stdout: ""` and `stderr: "1.3.0"`. I was unable to replicate this behavior in a shell, so I updated `execFile` to use "shell" mode by default, which solves the issue.

This seems to be specific to the MSYS2 package of meson, since using the pip version runs fine. However, since this is such a simple change that should not affect functionality, I thought I'd submit it.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -25,7 +25,12 @@ export async function exec(command: string, args: string[], options: cp.ExecFile
   });
 }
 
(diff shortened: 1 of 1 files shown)


tristan957:
Looks like there is a prettier issue.

aidenfoxx:
Sorry about that. Should be solved.

PR #205, opened 2023-11-28
Make sure rust-analyser uses rust-project.json Meson generates

xclaesse:
This is very similar case than compile_commands.json, Meson generates a file in builddir that some extensions should be using. It makes static analysers work out of the box instead of requiring user to configure paths themself.

We only need to watch for file creation, it was previously wrongly watching for file changes for legacy reasons (we used to patch compile_commands.json).

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,12 +6,12 @@ import {
   extensionConfiguration,
   extensionConfigurationSet,
   genEnvFile,
-  useCompileCommands,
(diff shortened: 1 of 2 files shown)


xclaesse:
This is a first step, there is probably more to be done. See discussion in https://github.com/rust-lang/rust-analyzer/issues/15963.

tristan957:
This looks good to me.

Issue #181, opened 2023-11-13
Meson isn't picked up from the `PATH`


paveloom:
I'm using [Nix](https://nixos.org), [`direnv`](https://github.com/direnv/direnv), and [`nix-direnv`](https://github.com/nix-community/nix-direnv) to setup a development environment. I'm using the [`direnv` VSCode extension](https://github.com/direnv/direnv-vscode) to load that environment in VSCode. This extension can't find Meson that is provided by the environment:

```
The terminal process failed to launch: Path to shell executable "meson" does not exist.
```

 I think that's because you're using [`ProcessExecution`](https://code.visualstudio.com/api/references/vscode-api#ProcessExecution) instead of [`ShellExecution`](https://code.visualstudio.com/api/references/vscode-api#ShellExecution). I poked around and found that you made a switch to `ProcessExecution` in https://github.com/mesonbuild/vscode-meson/pull/47 to avoid shell escaping issues. Are those issues still relevant? And if so, won't the second constructor for `ShellExecution` bypass them?

tristan957:
@lukester1975 can you check this out?

lukester1975:
Erm, dunno. Looks like it's all changed since those mods by me! 

I did mention `ShellQuoting` in that issue, but sadly I don't recall if that means I tried the second ctor and still had issues or what. The vscode type defs were old at the time but that second ctor was still there AFAICT. Sorry, not much help.

That said, FWIW, I just  `s/ProcessExecution/ShellExecution/` on main and it still worked for the couple of things I threw at it. But day-to-day I'm still on my old multiple build dir branch, so that change won't get much more testing from me!

Maybe it's best for @a paveloom to try running with those mods and see how it goes; seems a bit silly me making a PR for it?

tristan957:
If you don't want to PR it, I can. I'll take your experience as an endorsement for merging. When do you think you'll get your multiple build directory branch up for PR? I'd be really likely to merge it given you run it all the time.

lukester1975:
OK #186. As long as I don't get blamed...

R.e. multiple build dirs - lord knows. So much has changed and cursorily it looks like more stuff assumes a single build dir. It would be a complete rewrite I imagine, so probably unlikely until either a) my old branch no longer works with current vscode or b) there's something so good in the latest meson plugin that I want both bits of functionality...

Maybe I could rebase some of the trivial stuff one day and at least make a start that way.

Cheers

tristan957:
No problem. Thanks for looking.

Issue #203, opened 2023-11-28
Path issue when using cygwin as vscode integrated console on Windows


ZZH-Finzlize:
I'm using cygwin as vscode integrated console, but I found when I update this extension to version 1.8.1, it can't reconfigure my project any more, then I back to version 1.8.0. and that version has no problem.

v1.8.1:
![image](https://github.com/mesonbuild/vscode-meson/assets/33342565/a135ab06-47a6-4ee6-8ce7-2c9ad0fb7592)

v1.8.0:
![image](https://github.com/mesonbuild/vscode-meson/assets/33342565/e1f81cf7-9d30-465e-b016-8667eb7ee6e4)

tristan957:
Could you bisect the commit which broke it for you? I suspect that it might be https://github.com/mesonbuild/vscode-meson/commit/5a103c462f5a9d4e17bd9184e5b4fcfff3e1495c.

tristan957:
Does https://github.com/mesonbuild/vscode-meson/commit/3946f91725743a3b3ab0b0c37bb9368de8c82de1 work for you?

ZZH-Finzlize:
I guess https://github.com/mesonbuild/vscode-meson/commit/5a103c462f5a9d4e17bd9184e5b4fcfff3e1495c should be the reason, because of the command is different between two version, in version 1.8.0 it is "meson.exe" but in 1.8.1 it is "bash -c meson ....."

tristan957:
What does Meson resolve to in the broken command?

ZZH-Finzlize:
I think this problem is nothing to do with the Meson, Meson raise a error says "ERROR: Neither source directory 'd:workprojw3000stm32onboard-V3builddir' nor build directory 'd:workprojw3000stm32onboard-V3' exist.", clearly the "\" chars in the path are wrapped by "bash" as Escape character, bash is a linux shell so it accept Unix style path.

So could we add a setting to select the "ProcessExecution" or "ShellExecution"?

lukester1975:
Maybe https://github.com/lukester1975/vscode-meson/tree/more-shell-fixes ?

I successfully expunged cygwin from my life a few years ago so haven't checked it and I refuse to go back... OK on Linux/bash, Windows/cmd and Windows/git bash for me though.

There are the `ShellQuoting` options for each argument if not. Maybe one does the trick.

tristan957:
What you are proposing looks reasonable

tristan957:
[...]

PR #186, opened 2023-11-15
ProcessExecution -> ShellExecution to fix #181.

lukester1975:
Use at your own risk...

PR #206, opened 2023-11-28
Resolves mesonPath to program absolute path

xclaesse:
ProcessExecution() does not lookup into PATH.
    
Fixes: #181

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,6 +2,8 @@ import * as fs from "fs";
 import * as path from "path";
 import * as cp from "child_process";
 import * as vscode from "vscode";
+import * as which from "which";
+
 import { createHash, BinaryLike } from "crypto";
 import { ExtensionConfiguration, Target } from "./types";
 import { getMesonBuildOptions } from "./introspection";
@@ -219,4 +221,8 @@ export function whenFileExists(ctx: vscode.ExtensionContext, file: string, liste
   if (fs.existsSync(file)) {
     listener();
   }
+}
+
+export function mesonProgram(): string {
+  return which.sync(extensionConfiguration("mesonPath"));
 }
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -220,3 +222,7 @@ export function whenFileExists(ctx: vscode.ExtensionContext, file: string, liste
     listener();
   }
 }
+
+export function mesonProgram(): string {
+  return which.sync(extensionConfiguration("mesonPath"));
```
tristan957, line 227: Can you make this a variable instead? Otherwise, we are calling this a lot.

tristan957:
@ZZH-Finzlize can you also try this PR?

ZZH-Finzlize:
@tristan957 Yes, I tried this PR, it can solve #203 as well, and I noticed that this PR will call meson program by absolute path

Test result:
image

lukester1975:
Hmm. `ProcessExecution` does look in the path though. Nothing would have ever worked otherwise (the default for `mesonPath` is just `meson`, after all)...

In fact, trying direnv and before any changes related to #181, it picks up meson from `.envrc` just fine for me. Maybe the initial issue from #181 was just casued by a misconfiguration?

paveloom:
There is hardly anything to misconfigure. I checked and even commented on https://github.com/mesonbuild/vscode-meson/pull/186.

3e9cd2317d3be9a0fa3a77cb801c8b89da3ddd7a:

https://github.com/mesonbuild/vscode-meson/assets/49961859/aaa32bb7-1e46-4a22-84cb-086d92a91e20

5a103c462f5a9d4e17bd9184e5b4fcfff3e1495c:

https://github.com/mesonbuild/vscode-meson/assets/49961859/499a5ffa-cc13-4f68-b83b-28f44bf1f64d

lukester1975:
I guess I may have run vscode from the already direnved environment such that the PATH was correct, without really needing the direnv plugin to do anything. Maybe `ProcessExecution` only sees the pre-direnved environment's PATH...

Anyway no matter since this latest iteration does the job, was just more curious about what was going on!

PR #208, opened 2023-11-29
Allow root meson.build files to be symlinks

### LINKED ISSUE #207: Extension ignores root meson.build if it's a symlink
My root `meson.build` for a project is a symlink and the extension asks me to select one of the definitions from subprojects. The problem goes away if I replace the root `meson.build` with a regular file.

Likely caused by #195

Extension version: v1.19.0

tristan957:
Fixes #207

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -198,7 +198,7 @@ export async function mesonRootDirs(): Promise<string[]> {
     let hasMesonFile: boolean = false;
     let subdirs: vscode.Uri[] = [];
     for (const [name, type] of await vscode.workspace.fs.readDirectory(d)) {
-      if (type === vscode.FileType.File && name == "meson.build") {
+      if ((type === vscode.FileType.File || type === vscode.FileType.SymbolicLink) && name == "meson.build") {
         rootDirs.push(d.fsPath);
         hasMesonFile = true;
         break;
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -198,7 +198,7 @@ export async function mesonRootDirs(): Promise<string[]> {
     let hasMesonFile: boolean = false;
     let subdirs: vscode.Uri[] = [];
     for (const [name, type] of await vscode.workspace.fs.readDirectory(d)) {
-      if (type === vscode.FileType.File && name == "meson.build") {
+      if ((type === vscode.FileType.File || type === vscode.FileType.SymbolicLink) && name == "meson.build") {
```
andrei8l, line 201:
```suggestion
      if ((type === vscode.FileType.File || type === (vscode.FileType.File | vscode.FileType.SymbolicLink)) && name == "meson.build") {
```
AFAICT this is the correct check. Based on [this sample](https://github.com/microsoft/vscode-extension-samples/blob/main/nodefs-provider-sample/src/extension.ts#L267), and the debugger showing type is 65 for symlinked files.

tristan957: Makes sense. Can you submit a PR? I'll cook up a release tomorrow. Sorry!

PR #209, opened 2023-11-29
utils: fix symlink handling in mesonRootDirs

andrei8l:
Bitwise check for File or Directory type to properly handle symlinks of both

* Follow up for: #208

Testing:
- symlinked root `meson.build` gets picked up automatically
- symlinked directory with `meson.build` appears in the `Select configuration to use` prompt if there's no root `meson.build`
- regular files and directories still work
- (symlinked) directory named `meson.build` doesn't get picked up as root configuration

Based on [this sample](https://github.com/microsoft/vscode-extension-samples/blob/main/nodefs-provider-sample/src/extension.ts#L267).


tristan957:
Looks like Prettier is failing for the PR. Can you check on that?

Issue #9, opened 2019-09-09
Custom environment variables for meson configuration


MathiasMagnus:
Meson requires `PKG_CONFIG_PATH` to find some dependencies, and even though I setup my PS profile to include this var, I cannot set it as the default shell, as it breaks many Linux tooling. Currently I can't seem to find a way to tell the extension to include extra env vars during the configuration step.

tristan957:
This would probably be a pretty small PR if someone wanted to pick this up. I agree this would be a good nice to have.

xclaesse:
Note that in the case of `PKG_CONFIG_PATH` you can instead add `pkg_config_path=/foo` configure option.

ptomato:
I'd like to second this! My use case is a VSCodium session SSH'd into a Fedora Toolbox container. I'd like to be able to set the `DISPLAY` and `DBUS_SESSION_BUS_ADDRESS` environment variables when running Meson tests.

(For actually running the build product, you can set these environment variables in `launch.json`, but running tests doesn't take its environment variables from there. Also, I might expect it to respect the environment variables in the setting `terminal.integrated.env.linux`.)

I'd love to contribute this feature myself, if someone could give me a pointer on where to start. (I'm new to VSCode/VSCodium in general.) Would it be as simple as adding a configuration object to the `exec()` call in https://github.com/mesonbuild/vscode-meson/blob/10eb43d30a6d74c841abefca2597a33a900cf77e/src/tests.ts#L42?

tristan957:
Hello Mr. gjs :)

If you look at that exec() function, it is using [`child_process.execFile`](https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback) behind the scenes. We just wrap it into a promise. You can see that there is an env variable located there. We don't spawn the process in an integrated terminal, so that setting would not work for us.

ptomato:
Thanks very much! :smile: 

What would be a good place for the setting to live at, then? I'm not too familiar yet with VSCode so I don't really have a sense of this. Would it be a setting specific to the vscode-meson extension?

tristan957:
It should go in the package.json file. Should be fairly obvious where to put it!

PR #215, opened 2024-01-04
Environment variables for setup and test tasks

ptomato:
This adds two options, `configureEnvironment` and `testEnvironment`, allowing users to set particular environment variables during `meson setup` and `meson test`.

Closes: #9

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,14 @@
           "default": [],
           "description": "Specify the list of options used for Meson setup."
         },
+        "mesonbuild.configureEnvironment": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          },
+          "default": {},
+          "description": "Specify the list of additional environment variables used for setup."
+        },
         "mesonbuild.setupOptions": {
           "type": "array",
           "default": [],
--- a/package.json
+++ b/package.json
@@ -122,12 +122,12 @@
         "mesonbuild.buildFolder": {
           "type": "string",
           "default": "builddir",
-          "description": "Specify in which folder Meson build configure and build the project. Changing this value will reload the VS Code window."
+          "description": "Specify in which folder Meson configures and builds the project. Changing this value will reload the VS Code window."
         },
         "mesonbuild.configureOptions": {
           "type": "array",
           "default": [],
-          "description": "Specify the list of options used for Meson setup."
+          "description": "Specify the list of options used for setup."
         },
         "mesonbuild.configureEnvironment": {
           "type": "object",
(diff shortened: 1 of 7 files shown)


ptomato:
I'm not really sure if this is the right way to do it, so please be zealous in reviewing it :smile:

PterX:
any update here?

tristan957:
Yeah... that's my bad. Let me get to it this week.

PterX:
> Yeah... that's my bad. Let me get to it this week.

thx

src/utils.ts
```diff
@@ -15,7 +15,15 @@ export interface ExecResult {
...
 
-export async function exec(command: string, args: string[], options: cp.ExecFileOptions = { shell: true }) {
+export async function exec(
+  command: string,
+  args: string[],
+  extraEnv: { [key: string]: string } | undefined = undefined,
+  options: cp.ExecFileOptions = { shell: true },
+) {
+  if (extraEnv) {
+    options.env = { ...(options.env ?? process.env), ...extraEnv };
```
tristan957, line 25: Maybe I don't understand the node API here, if we pass an environment object, does it not also inherit the parent environment?

package.json
```diff
@@ -146,6 +146,14 @@
...
           "type": "array",
           "default": [],
+          "mesonbuild.testEnvironment": {
+            "type": "object",
+            "additionalProperties": {
+              "type": "string"
+            },
+            "default": {},
+            "description": "Specify the list of additional environment variables used for running tests."
+          },
```
tristan957, line 156: This looks nested within mesonbuild.testOptions, which is incorrect

tristan957:
Just a few comments, and then it should be good to go. It would also be useful if you could add an entry in the CHANGELOG.md file. Otherwise, I can get to it after merge.

package.json
```diff
@@ -129,6 +129,14 @@
           "default": [],
           "description": "Specify the list of options used for Meson setup."
         },
+        "mesonbuild.configureEnvironment": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          },
+          "default": {},
+          "description": "Specify the list of additional environment variables used for Meson setup."
```
tristan957, line 138: I would remove "Meson" here.

src/utils.ts
ptomato: I'm not 100% sure. The documentation says, about the `env` option:

> `env`: Environment key-value pairs. Default: `process.env`.

So because the default is `process.env` and not `{}`, I'd guess that `options.env` replaces the parent environment. (That also makes sense because otherwise, you couldn't delete environment variables.)

package.json
ptomato: I've also added a new commit to remove it elsewhere.

tristan957 (approved):

PR #218, opened 2024-02-02
Add a Meson cpptools configProvider

deribaucourt:
# Description from the commit message

    Untill now the meson extension provided C++ linting through
    compile_commands.json. This would not configure:
     - New file that haven't been compiled yet
     - External libraries that are not part of the project (when browsing/debugging)
    This was especially problematic when using a cross-compilation toolchain
    since none of the default host headers should be used.
    
    This adds a custom configuration provider for the cpptools extension. It
    introspects the compiler settings of the project to configure linting
    and browsing. Currently, file-specific configurations are not supported.
    Some other advanced meson options might also not be supported. However
    it provides a strong experience for cross-compilation and external
    library support.
    
    Here is an example settings.json configuration to activate this provider:
    ```json
    {
      "C_Cpp.default.configurationProvider": "mesonbuild.mesonbuild"
    }
    ```

# Testing the configuration

One can quickly test the generated configuration by removing the `compile_commands.json` and browsing C/C++ files. Most include paths and compiler flags should still appear effective. You can for instance add a dependency to a nonstandard library and the `#include`  statement should not produce error squiggles.
Either remove the option in the .vscode/settings.json file at runtime, or use a patch
```diff
diff --git a/src/extension.ts b/src/extension.ts
index dedc767..a51ff61 100644
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -142,7 +142,7 @@ export async function activate(ctx: vscode.ExtensionContext) {
   whenFileExists(ctx, compileCommandsFile, async () => {
     if (shouldModifySetting("ms-vscode.cpptools")) {
       const conf = vscode.workspace.getConfiguration("C_Cpp");
-      conf.update("default.compileCommands", compileCommandsFile, vscode.ConfigurationTarget.Workspace);
+      // conf.update("default.compileCommands", compileCommandsFile, vscode.ConfigurationTarget.Workspace);
     }
   });
```


src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export class CpptoolsProvider implements cpptools.CustomConfigurationProvider {
+  cppToolsAPI?: cpptools.CppToolsApi;
+  private buildDir: string = "";
+
+  constructor(buildDir: string) {
+    this.buildDir = buildDir;
+  }
+
+  name = "Meson build";
```
tristan957, line 17: Let's make this "Meson Build"

src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
+import path = require("path");
+import * as vscode from "vscode";
+import * as cpptools from "vscode-cpptools";
+import { getMesonBuildOptions, getMesonCompilers, getMesonDependencies } from "./introspection";
+var fs = require("fs");
+
+export class CpptoolsProvider implements cpptools.CustomConfigurationProvider {
+  cppToolsAPI?: cpptools.CppToolsApi;
+  private buildDir: string = "";
```
tristan957, line 11: Remove the default value

src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
+import path = require("path");
+import * as vscode from "vscode";
+import * as cpptools from "vscode-cpptools";
+import { getMesonBuildOptions, getMesonCompilers, getMesonDependencies } from "./introspection";
+var fs = require("fs");
```
tristan957, line 7: Don't use require. Use import.

src/cpptoolsconfigprovider.ts
tristan957: Make these private

src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    let machine = "";
+    const buildOptions = await getMesonBuildOptions(this.buildDir);
+    for (let option of buildOptions) {
+      if (option.name === "cpp_std") {
+        browseConfig = Object.assign({}, browseConfig, { standard: option.value });
+        machine = option.machine;
+      }
+    }
```
tristan957: Is it possible for  us to not have a machine?

src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
...
+  ): Promise<cpptools.WorkspaceBrowseConfiguration | null> {
+    let browseConfig: cpptools.WorkspaceBrowseConfiguration = {
+      browsePath: [],
+      compilerPath: "",
+      compilerArgs: [],
+      standard: "c++17",
+    };
+
+    const dependencies = await getMesonDependencies(this.buildDir);
+    for (let dep of dependencies) {
```
tristan957: Use const instead of let

src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    return browseConfig;
+  }
+
+  // We only handle project-wide configurations.
+  canProvideBrowseConfigurationsPerFolder(token?: vscode.CancellationToken | undefined): Thenable<boolean> {
+    return Promise.resolve(false);
+  }
+  async provideFolderBrowseConfiguration(
```
tristan957, line 112: Whitespace around functions

src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
+import path = require("path");
+import * as vscode from "vscode";
+import * as cpptools from "vscode-cpptools";
+import { getMesonBuildOptions, getMesonCompilers, getMesonDependencies } from "./introspection";
+var fs = require("fs");
+
+export class CpptoolsProvider implements cpptools.CustomConfigurationProvider {
+  cppToolsAPI?: cpptools.CppToolsApi;
```
tristan957, line 10: Make this private?

src/extension.ts
```diff
@@ -69,6 +71,8 @@ export async function activate(ctx: vscode.ExtensionContext) {
   const buildDir = getBuildDirectory(sourceDir);
   workspaceState.update("mesonbuild.buildDir", buildDir);
   workspaceState.update("mesonbuild.sourceDir", sourceDir);
+  cpptools = new CpptoolsProvider(buildDir);
+  registerCppToolsProvider(ctx, cpptools);
```
tristan957, line 76: Should we only do this if it is a C or C++ project?

src/cpptoolsconfigprovider.ts
```diff
@@ -0,0 +1,120 @@
+import path = require("path");
```
tristan957, line 1: We should document this in the README.md somewhere.

tristan957 (commented):
It seems to me like a few of the commits could be squashed together.

src/cpptoolsconfigprovider.ts
deribaucourt: Renaming to "Meson Build". However, name and extensionId have to remain public because this class `implements cpptools.CustomConfigurationProvider`

src/cpptoolsconfigprovider.ts
deribaucourt: Done

src/cpptoolsconfigprovider.ts
deribaucourt: Done

src/cpptoolsconfigprovider.ts
deribaucourt: Done

src/cpptoolsconfigprovider.ts
deribaucourt: Done

src/extension.ts
deribaucourt: I don't think it matters. If the user does not add the `"C_Cpp.default.configurationProvider": "mesonbuild.mesonbuild"` setting, then non of the configuration methods will be called. And they do, this only affects C/C++ files detected by the cpptools extension.

src/cpptoolsconfigprovider.ts
deribaucourt: Done

src/cpptoolsconfigprovider.ts
deribaucourt: Done

src/cpptoolsconfigprovider.ts
tristan957: Ah Thanks for the explanation!

src/cpptoolsconfigprovider.ts
deribaucourt: I tested the simple [meson test-case gtk-doc](https://github.com/mesonbuild/meson/tree/master/test%20cases/frameworks/10%20gtk-doc) with a native build. A machine was also defined ("host"). So it looks like meson always defines a machine, even with empty/default options.

tristan957: Let's add an assert and make `machine` `string | undefined`

deribaucourt:
Thanks a lot for your review. I made the recommended changes. Also added 2 commits for default configurations that improved the experience when I tested a native simple meson test-case.

src/cpptoolsconfigprovider.ts
deribaucourt: Alright! Added a commit to handle an undefined machine. I used an if rather than assert because we are still able to provide a browsePath without this option.

tristan957 (approved):

PR #219, opened 2024-03-15
Add mesonlsp

JCWasmx86:
This adds the C++-Rewrite of `Swift-MesonLSP` as a separate `mesonlsp` server. `Swift-MesonLSP` users will be upgraded to `mesonlsp`. If there are regressions, users can fallback to `Swift-MesonLSP-legacy`.

Issues:
- Switching language at runtime doesn't work.
- Not sure what to do with the options (Copy paste a old?)
- Not sure whether the changes to yarn.lock are needed
- I'm not sure how to make this code less hacky

And important: Before this language server is rolled out, I would like to ask adventurous people for feedback. Sure mesonlsp is perfect for my usage of meson and the usage of tristan957 (Thanks for testing!), but I think it's important to first roll it out to e.g. 10 users (That have installed `mesonlsp` by themselves and then continue with the entire rollout


tristan957:
I would recommend just dropping the old server. VSCode allows pinning an extension if people really want it

tristan957:
I still think that is where pinning comes into play

tristan957 (approved):

PR #231, opened 2024-05-06
Fix failure to run executable targets on Windows

### LINKED ISSUE #228: Run executable fails on Windows
I'm new to Meson, so I first tried to create just a simple binary project and then run it, but was faced with a problem: the task responsible for running executable targets fails unconditionally with the message "A native exception occurred during launch (Cannot create process, error code: 87)." in VSCode terminal. After some investigation, I was able to fix this with a few lines in the `genEnvFile()` function:
```typescript
 for (const i of data.toString().split(/\r?\n/)) {
    // skip empty lines
    if (!i) {
        continue;
    }

    // Poor man's i.split("=", 1), JS won't return part after first equal sign.
    // Value is quoted, remove first and last " char and also possible \r ending.
    const index = i.indexOf("=");
    const key = i.substring(0, index);
    const value = i.slice(index + 2, -1);
    _envDict[key] = value;
  }
```
Alternatively, `filter(i => i)` on the list returned by `split()` will do the same.

cerfical:
Fixes #228

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -168,7 +168,10 @@ export async function genEnvFile(buildDir: string) {
   // Load into a dict because vscode.ProcessExecution() does not support envFile.
   _envDict = {};
   const data = await fs.promises.readFile(envfile);
-  for (const i of data.toString().split(/\r?\n/)) {
+  for (const i of data
+    .toString()
+    .split(/\r?\n/)
+    .filter((i) => i)) {
     // Poor man's i.split("=", 1), JS won't return part after first equal sign.
     // Value is quoted, remove first and last " char and also possible \r ending.
     const index = i.indexOf("=");


tristan957:
Could you describe how this fixes the issue? Thanks for your contribution

cerfical:
Well, the code below parses the environment file `meson-vscode.env`:

https://github.com/mesonbuild/vscode-meson/blob/3f4ee1e3e5235bfc32995feab86e0646febcacd5/src/utils.ts#L169-L178

I haven't looked into how exactly the file gets its contents, but I guess that it must contain some kind of environment variables defined as `<key>=<value>` pairs. Later, the environment is used to initialize a new process responsible for a run task:

https://github.com/mesonbuild/vscode-meson/blob/3f4ee1e3e5235bfc32995feab86e0646febcacd5/src/tasks.ts#L47-L50

The problem occurs because `envFile` can contain an empty line (specifically at the end of file), which causes an empty key with an empty value to be inserted into the resulting dict. The implementation then calls a system API to create a process and fails. So for a simple solution, we can just ignore empty lines by `filter()`'ing them out. Or we could redesign some things, and that requires extra work for such a small issue, so...

PR #237, opened 2024-05-31
Fix tests with spaces in their names not executing when ran from the testing ui

apache-hb:
Previously running a test which contained a space in its name, ie `command line parsing` would result in the test runner printing 
```
ERROR: *:command test name does not match any test
```
This encloses the test name in quotes so its parsed correctly on the command line.

### DIFF
--- a/src/tests.ts
+++ b/src/tests.ts
@@ -41,7 +41,7 @@ export async function testRunHandler(
     try {
       await exec(
         extensionConfiguration("mesonPath"),
-        ["test", "-C", buildDir, "--print-errorlog", test.id],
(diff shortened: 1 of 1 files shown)


apache-hb:
> Can you use a real commit message? 

Fixed

> Can you confirm that this actually works.


I tested it by running tests from the vscode test provider and having them pass rather than error out with the error message mentioned above.
image

> Upon first inspection, I don't really see why this would work.

This works because `child_process.execFile` - at least on windows - concatenates all arguments together naïvely before handing them off to `CreateProcess`. Adding the quotes makes meson/python parse the test name as a single argument rather than multiple.

tristan957:
Do we have the same problem if `buildDir` has a space in it?

PR #239, opened 2024-06-05
Limit the depth of the meson.build search

### LINKED ISSUE #234: Extention doesn't activate if meson.build is not in the workspace root
Because of the Perforce specifics, I can't have `meson.build` in the workspace root. My `meson.build` is at `${workspaceFolder}/src`. In this configuration, the Meson extension doesn't even start. There are no Meson commands, and there is no Meson section in the OUTPUT console. In the Extension Host's section of the OUTPUT console, I see
```
2024-05-16 20:22:01.728 [info] ExtensionService#_doActivateExtension mesonbuild.mesonbuild, startup: false, activationEvent: 'onLanguage:meson'
```
That's it. I tried to cleanly remove the extension and reinstall it. No effect.

Meson extension version is 1.24.0.
VS Code 1.89.1.
OS Linux.

slavaandrejev:
Searching the entire workspace can be very costly, especially if the workspace is on a network drive. This change makes the extension assume that meson.build is located either in the workspace's root or in any directory two levels down from the root.

Potentially fixes #234.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,9 @@
     "Programming Languages"
   ],
   "activationEvents": [
-    "workspaceContains:**/meson.build",
+    "workspaceContains:meson.build",
+    "workspaceContains:*/meson.build",
+    "workspaceContains:*/*/meson.build",
     "onDebugDynamicConfigurations",
     "onDebugDynamicConfigurations:cppdbg",
     "onDebugDynamicConfigurations:lldb"


tristan957:
Could you try this locally on your monorepo to see if it fixes your issue?

tristan957:
Note that I am going to make a bugfix release today, so if you want this in, please get back to me

intractabilis:
> Note that I am going to make a bugfix release today, so if you want this in, please get back to me

I don't know what I am supposed to do. I don't know TypeScript, and I've never developed VS Code extensions. I have zero idea how to verify Slava's PR. So, I guess it's ok. I got used to living with this bug.

tristan957:
Open this repo in VSCode. Checkout slava's branch. Then launch a debug session. That will open a new vscode window, which you can then use to open your huge monorepo.

intractabilis:
> Open this repo in VSCode. Checkout slava's branch. Then launch a debug session. That will open a new vscode window, which you can then use to open your huge monorepo.

It says it needs `yarn`, whatever that might be. It's a corporate laptop, so I can't install whatever I want here.

tristan957:
Well I don't want to merge this until I get confirmation that it fixes the issue, so until that point that it is confirmed, I guess this will sit in limbo.

tristan957:
https://classic.yarnpkg.com/lang/en/docs/install/

slavaandrejev:
IMHO, limiting the search depth is the right thing to do, regardless. When I worked at NVIDIA, my client space could have a dozen gigabytes of source code. It probably would have taken an hour for VS Code to fumble through all the files, even on a local drive. Also, IMHO, asking uninitiated users to install development tools is counterproductive.

However, I understand—it's your kitchen, your rules. I am okay with this PR sitting in limbo.

@intractabilis You may consider other Meson extensions. The feature you need most is syntax highlighting. Meson's setup and compile phases are easily configurable in `tasks.json`.

tristan957:
I think people who come to the table with fairly unique use cases need to be prepared to be able to test. If I can't repro, I can't help, but I'll just go ahead and merge this.

intractabilis:
> I think people who come to the table with fairly unique use cases need to be prepared to be able to test. If I can't repro, I can't help, but I'll just go ahead and merge this.

You certainly have the full right to set up such a rule. Moreover, I never objected to testing. It's just that your proposed unusual way of testing was a no-go. I filed several bugs against VS Code itself. Microsoft has never asked me to compile VS Code to see if their fix worked. Their approach is: Hey, the fix made its way to the release X; please check if it works for you.

PR #241, opened 2024-06-17
Pass environment to the testing debugger

### LINKED ISSUE #240: Environment not working for debug test framework
Thank you for the great extension!

Steps:
1. I am using vscode-meson v1.24.0 with a C/C++ project
1. Select VSCode's "Testing" feature from the Activity Bar
2. Select "Debug Test"

The environment is not passed to the debugger. The environment is passed OK for "Run Test".

image

cskeogh:
The C++ debugger requires "environment" key in the format
[ { "name": "config", "value": "Debug" } ]
https://code.visualstudio.com/docs/cpp/launch-json-reference

Note: I think [Javascript](https://code.visualstudio.com/docs/editor/debugging) uses 'env'. [Python](https://code.visualstudio.com/docs/python/debugging#_env) uses 'env'. Annoying. This patch doesn't account for Javascript or Python. Not sure if that is a problem.

Fixes #240

### DIFF
--- a/src/tests.ts
+++ b/src/tests.ts
@@ -1,6 +1,6 @@
 import * as vscode from "vscode";
 import { ExecResult, exec, extensionConfiguration } from "./utils";
-import { Tests } from "./types";
+import { Tests, DebugEnvironmentConfiguration } from "./types";
 import { getMesonTests, getMesonTargets } from "./introspection";
 import { workspaceState } from "./extension";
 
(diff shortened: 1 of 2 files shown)


tristan957:
Actually, maybe it would be better to special case it in a if statement?

cskeogh:
> Actually, maybe it would be better to special case it in a if statement?
I added the if statement 👍

tristan957:
Can you squash the commits? I'll make a new release today.

cskeogh:
> Can you squash the commits? I'll make a new release today.

Squashed 👍

PR #247, opened 2024-06-30
Categorise test cases into test suite(s)

### LINKED ISSUE #245: Categorise test cases into test suite(s)
Thank you for the great extension!

I think the vscode-meson should categorise test cases into test suite(s), if exists.

Steps:

1. I am using vscode-meson v1.24.0 with a C/C++ project (gtk)
2. Select VSCode's "Testing" feature from the Activity Bar
3. Observe a long list of test cases. The gtk project categorises the Meson test cases into Meson test suite(s). I think vscode&#8209;meson should categorise too.

Screenshot 2024-06-30 121436

cskeogh:
Meson test cases may belong to test suite(s). If so, in VSCode's "Testing" feature from the Activity Bar, add the test suite(s) as a parent item and add the test cases as children.

The user has the choice to run all tests belonging to a test suite. Will pass the test suite to the meson test command --suite. The option --suite is needed to run test cases that the meson config purposely excludes (for example, test suite "flaky").

Fixes #245

Quote the "test case" in the meson command, so can run test cases with a space in the name. The quotes are needed because `cp.ExecFileOptions` sets `shell=true`.

Fixes #256

### DIFF
--- a/src/tests.ts
+++ b/src/tests.ts
@@ -50,16 +50,16 @@ function deleteTestsFromControllerNotVisited(
   controller: vscode.TestController,
   testsVisited: WeakMap<vscode.TestItem, boolean>,
 ) {
-  for (const [test_id, test] of controller.items) {
+  for (const [testId, test] of controller.items) {
     if (testsVisited.get(test) == undefined) {
-      for (const [child_id] of test.children) {
-        test.children.delete(child_id);
+      for (const [ChildId] of test.children) {
+        test.children.delete(ChildId);
       }
-      controller.items.delete(test_id);
+      controller.items.delete(testId);
     } else {
-      for (const [child_id, child] of test.children) {
+      for (const [ChildId, child] of test.children) {
         if (testsVisited.get(child) == undefined) {
-          test.children.delete(child_id);
+          test.children.delete(ChildId);
         }
       }
     }
@@ -67,7 +67,7 @@ function deleteTestsFromControllerNotVisited(
 }
 
 export async function rebuildTests(controller: vscode.TestController) {
-  let tests = await getMesonTests(workspaceState.get<string>("mesonbuild.buildDir")!);
+  const tests = await getMesonTests(workspaceState.get<string>("mesonbuild.buildDir")!);
   const testsVisited = addMesonTestsToController(controller, tests);
   if (testsVisited != null) {
     deleteTestsFromControllerNotVisited(controller, testsVisited);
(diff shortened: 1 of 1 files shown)


cskeogh:
Using the gtk project, showing the test suite(s):

Screenshot 2024-06-30 124213

tristan957:
Thanks for the PR. I will try to test and review it this week.

src/tests.ts
```diff
@@ -4,18 +4,73 @@ import { Tests, DebugEnvironmentConfiguration } from "./types";
...
+}
 
-  controller.items.forEach((item) => {
-    if (!tests.some((test) => item.id == test.name)) {
-      controller.items.delete(item.id);
+function deleteTestsFromControllerNotVisited(
+  controller: vscode.TestController,
+  testsVisited: WeakMap<vscode.TestItem, boolean>,
+) {
+  for (const [test_id, test] of controller.items) {
```
tristan957, line 53: Can you make variable names camelCase?

src/tests.ts
```diff
@@ -4,18 +4,73 @@ import { Tests, DebugEnvironmentConfiguration } from "./types";
...
     }
-  });
+  }
+}
 
-  for (let testDescr of tests) {
-    let testItem = controller.createTestItem(testDescr.name, testDescr.name);
-    controller.items.add(testItem);
+export async function rebuildTests(controller: vscode.TestController) {
+  let tests = await getMesonTests(workspaceState.get<string>("mesonbuild.buildDir")!);
```
tristan957, line 70: const

tristan957:
What do you think of #237? Should I merge that one first?

cskeogh:
Updated the PR to fix the review comments.

Note the PR does have some limitations:

1. Gtk's test case "about state" is in 2 test suites "gtk+:a11y" and "gtk+:flaky". This PR shows "about state" correctly in both test suites, but they are separate instances of `TestItem`. Ideally, "about state" would be one instance of `TestItem`, so that the failure icon shows on both (see screenshot yellow highlight). But VSCode doesn't support one `TestItem` with 2 parents. I tried.
 Screenshot 2024-07-02 215050
 Screenshot 2024-07-02 215210

1.  The user can run the test suite, which will correctly run all test cases belonging to the test suite. But the children `TestItem` do not show the success / failure icons (see screenshot yellow highlight). Only the test suite shows success / failure. In the below example, I ran test suite "gtk+:gdk" and 7 test cases passed and 1 failed:
Screenshot 2024-07-02 220438
  ```
Ok:                 7   
Expected Fail:      0   
Fail:               1   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0 
   ```

tristan957:
Thanks for explaining the limitations. I see that there is this function. Do we have access to the test runs, so if someone runs a suite, we can properly annotate the test? https://vscode-api.js.org/interfaces/vscode.TestRun.html#passed

cskeogh:
I'm working on resolving the limitations. Closing this PR until I resolve.

tristan957:
Alright. Really appreciate you looking into this more!

PR #256, opened 2024-08-08
Improve debug configuration providers

ylatuya:
The following improves the debug configuration providers:

- Fix a bug in the creation of debug configurations for  `cppdbg` and `cppvsdbg`
- Register `cppdbg` and `cppvsdbg` providers separetly
- Only register the providers if there is a debugger extension installed.


ylatuya:
Here is an example of providing debug configuration for a debugger that doesn't exist:
Screenshot 2024-08-07 at 14 22 54

ylatuya:
I am not sure how to handle the case of a missing debugger extension:

1. Only register the provider if the extension exists
2. Always register the providers, but only provide debug configurations when the extension exists.

src/debug/index.ts
```diff
@@ -61,4 +62,41 @@ export abstract class MesonDebugConfigurationProvider implements vscode.DebugCon
...
+      program: target.filename[0],
+      args: [],
+      preLaunchTask: `Meson: Build ${targetName}`,
+    };
+
+    if (type === DebuggerType.cppdbg) {
+      let miMode;
+      if (
+        target.target_sources?.some((source) => source.compiler != null && ["cc", "clang"].includes(source.compiler[0]))
+      ) {
```
tristan957, line 83: Not sure I understand why cc means clang.

tristan957:
What line does the undefined come from?

ylatuya:
> What line does the undefined come from?

`undefined` comes from registering a debug configuration provider using `vscode.debug.registerDebugConfigurationProvider` with a debugger type that is unknown, for example registering it for the `lldb` debugger type when this debugger does not exists.

Debuggers are registered by extensions like CodeLLDB [here](https://github.com/vadimcn/codelldb/blob/master/package.json#L653 ) registering a new debugger with `type: lldb` and `label: LLDB`.

src/debug/index.ts
ylatuya: This is definitively a wrong assumption. Since it's hard to infer it, I suggest using the [defaults](https://code.visualstudio.com/docs/cpp/launch-json-reference#_customizing-gdb-or-lldb) based on the operating system. This could be customizable in the future with a new configuration property.

tristan957:
> undefined comes from registering a debug configuration provider using vscode.debug.registerDebugConfigurationProvider with a debugger type that is unknown, for example registering it for the lldb debugger type when this debugger does not exists.

Does this mean that we should check for the existence of the debugger before registering the provider?

ylatuya:
> > undefined comes from registering a debug configuration provider using vscode.debug.registerDebugConfigurationProvider with a debugger type that is unknown, for example registering it for the lldb debugger type when this debugger does not exists.
> 
> Does this mean that we should check for the existence of the debugger before registering the provider?

This is what this PR does here: https://github.com/mesonbuild/vscode-meson/pull/256/commits/fb4d6a80fe18d367470fff9b5fb7cb87013e28c9

My doubt is whether to do that check when registering the provider or when listing configurations in `MesonDebugConfigurationProvider.provideDebugConfigurations` for the corner case in which a user has the meson extension installed and initialized and they install the extension providing the debugger support afterwards, which would require restarting vscode. Supporting that corner would make the code less obvious, and I am unsure about the real benefits.

tristan957:
I don't think we need to solve the corner case. I like what you have. Are you ready for me to merge it?

ylatuya:
> I don't think we need to solve the corner case. I like what you have. Are you ready for me to merge it?

It's ready to be merged :) Thanks for the review

PR #260, opened 2024-09-21
support parallel test execution

Jannik2099:
This adds parallel test execution through the vscode TestController API. Sequential tests `test(..., is_parallel = false)` are respected.

Parallelism is set to the number of CPUs, if desired I could also add a config option.

I also fixed output formatting of test failures as `vscode.TestRun.appendOutput` always expects CLRF.


Jannik2099:
I couldn't help myself and just added test lookup functionality (right click on test -> Go to Test).

This leads to the source file that a test exe is built from. This only works for simple tests that are made up of one source file, but that should be the majority of tests.

tristan957:
@Jannik2099 do you think this PR is ready to go? Sorry it's taken me a long time to get to it

src/tests.ts
```diff
@@ -25,40 +55,78 @@ export async function testRunHandler(
...
+    const running_test = dispatchTest(test).finally(() => {
+      running_tests.splice(running_tests.indexOf(running_test), 1);
+    });
+
+    running_tests.push(running_test);
+
+    if (running_tests.length >= max_running) {
+      await Promise.race(running_tests);
     }
   }
```
tristan957, line 124: Should we provide an option limit the amount of parallelization

src/utils.ts
```diff
@@ -12,6 +12,7 @@ import { extensionPath, workspaceState } from "./extension";
 export interface ExecResult {
   stdout: string;
   stderr: string;
+  time: number; // runtime in milliseconds
```
tristan957, line 15: Maybe let's add a suffix to this member: `time_ms`

src/tests.ts
```diff
@@ -1,11 +1,40 @@
...
+  const test_exe = test.cmd.at(0);
+  if (!test_exe) {
+    return undefined;
+  }
+
+  // the meson target such that it is of meson type executable()
+  // and produces the binary that the test() executes.
+  const testDependencyTarget = targets.find((target) => {
+    const depend = test.depends.find((depend) => {
+      return depend == target.id && depend.endsWith("@exe");
```
tristan957, line 20: Hmm does the meson introspection stuff really not provide a better heuristic for determining whether a target is an exe?

src/tests.ts
```diff
@@ -1,11 +1,40 @@
...
-import { ExecResult, exec, extensionConfiguration } from "./utils";
-import { Tests, DebugEnvironmentConfiguration } from "./types";
+import { ExecResult, exec, extensionConfiguration, getTargetName } from "./utils";
+import { Targets, Test, Tests, DebugEnvironmentConfiguration } from "./types";
 import { getMesonTests, getMesonTargets } from "./introspection";
 import { workspaceState } from "./extension";
 
+// this is far from complete, but should suffice for the
+// "test is made of a single executable is made of a single source file" usecase.
+function findSourceOfTest(test: Test, targets: Targets) {
```
tristan957, line 10: Can you add an explicit return type

Jannik2099:
Sorry for the delay, I managed to break my finger which made interfacing with a keyboard... difficult.
Programming is a dangerous endeavour!

I addressed all your points.

tristan957:
I guess this slipped my mind when cutting the new release. Is there any way you could rebase on main, so I can get this for 1.28?

Jannik2099:
No worries, I rebased & tested that things still work as expected.

tristan957:
Thanks for your work. Will review next week.

Jannik2099:
ping?

package.json
```diff
@@ -158,6 +158,12 @@
           "default": {},
           "description": "Specify the list of additional environment variables used for running tests."
         },
+        "mesonbuild.testJobs": {
+          "type": "integer",
+          "default": -1,
+          "minimum": -1,
+          "description": "Specify the maximum number of tests executed in parallel. -1 for number of CPUs, 0 for unlimited."
+        },
```
tristan957, line 166: Instead of magic integers can we use strings to indicate unlimited and ncpus?

src/tests.ts
```diff
@@ -1,11 +1,40 @@
...
 import * as vscode from "vscode";
-import { ExecResult, exec, extensionConfiguration } from "./utils";
-import { Tests, DebugEnvironmentConfiguration } from "./types";
+import { ExecResult, exec, extensionConfiguration, getTargetName } from "./utils";
+import { Targets, Test, Tests, DebugEnvironmentConfiguration } from "./types";
 import { getMesonTests, getMesonTargets } from "./introspection";
 import { workspaceState } from "./extension";
 
+// this is far from complete, but should suffice for the
+// "test is made of a single executable is made of a single source file" usecase.
```
tristan957, line 9: Are you going to complete this TODO

src/tests.ts
```diff
@@ -25,39 +55,87 @@ export async function testRunHandler(
...
+    })!;
+    if (mesonTest.is_parallel) {
+      parallelTests.push(test);
+    } else {
+      sequentialTests.push(test);
+    }
+    // this way the total number of runs shows up from the beginning,
+    // instead of incrementing as individual runs finish
+    run.enqueued(test);
+  }
```
tristan957, line 76: Can you make this an arrow function?

const xyz = ...

src/tests.ts
```diff
@@ -25,39 +55,87 @@ export async function testRunHandler(
...
 ) {
   const run = controller.createTestRun(request, undefined, false);
-  const queue: vscode.TestItem[] = [];
+  const parallelTests: vscode.TestItem[] = [];
+  const sequentialTests: vscode.TestItem[] = [];
 
+  const buildDir = workspaceState.get<string>("mesonbuild.buildDir")!;
+  const mesonTests = await getMesonTests(buildDir);
+
+  function testAdder(test: vscode.TestItem) {
```
tristan957, line 64: Can you add a comment about what this function does?

src/tests.ts
```diff
@@ -25,39 +55,87 @@ export async function testRunHandler(
...
+          run.failed(test, new vscode.TestMessage(execResult.stderr), execResult.time_ms);
+        }
+      },
+    );
+  }
+
+  const running_tests: Promise<void>[] = [];
+  const max_running: number = (() => {
+    const jobs_config = extensionConfiguration("testJobs");
+    switch (jobs_config) {
```
tristan957, line 114: Please use camel case for naming

tristan957 (commented):
Some initial comments

package.json
Jannik2099: the "=0 means unlimited and =-1 means nproc" mimics semantics in ninja and most other tools with a `-j` option.
I don't have any strong feelings in either direction, but I don't think you can do "choice of two specific strings, or a number field" as a vscode setting?

src/tests.ts
Jannik2099: No.
The comment is about tests that consist of multiple TUs, or tests that are not native meson tests (see https://mesonbuild.com/Reference-manual_functions.html#test_protocol ) - I don't have a good idea on what projects using these options look like, as I personally only use single TU tests.
I'm not sure if there is a simple solution for these, and "single TU tests" should cover the majority of users.

Jannik2099:
converted back to draft as I discovered a minor issue - parallel runs could spawn parallel compile jobs.

package.json
tristan957: Your logic makes sense. Let's keep it.

src/tests.ts
```diff
@@ -1,11 +1,82 @@
...
+    ?.sources?.at(0);
+  return path ? vscode.Uri.file(path) : undefined;
+}
+
+/**
+ * ensures that the given test targets are up to date
+ * @param tests tests to rebuild
+ * @param buildDir meson buildDir
+ * @returns `ExecResult` of the `meson compile ...` invocation
+ */
```
tristan957, line 39: In all your comments (function descriptions, parameter descriptions), can you capital letters?

src/tests.ts
```diff
@@ -1,11 +1,82 @@
...
+
+function vsCodeToMeson(vsCodeTests: readonly vscode.TestItem[], mesonTests: Tests): Test[] {
+  return vsCodeTests.map((test) => {
+    return mesonTests.find((mesonTest) => {
+      return mesonTest.name == test.id;
+    })!;
+  });
+}
+
+export async function regenerateTests(controller: vscode.TestController) {
```
tristan957, line 76: Can you add some more function comments here please?

src/utils.ts
```diff
@@ -19,6 +19,7 @@ import { extensionPath, workspaceState } from "./extension";
 export interface ExecResult {
   stdout: string;
   stderr: string;
+  time_ms: number; // runtime in milliseconds
```
tristan957, line 22: timeMs, unless there was a reason you were keeping the underscore

src/utils.ts
```diff
@@ -32,11 +33,13 @@ export async function exec(
     options.env = { ...(options.env ?? process.env), ...extraEnv };
   }
   return new Promise<ExecResult>((resolve, reject) => {
+    let time_ms = Date.now();
```
tristan957, line 36: const

src/utils.ts
```diff
@@ -49,8 +52,10 @@ export async function execFeed(
   stdin: string,
 ) {
   return new Promise<ExecResult>((resolve) => {
+    let time = Date.now();
```
tristan957, line 55: const, also please make the variable the same name as the one above

tristan957 (commented):
Getting close!

Jannik2099:
ping?

tristan957:
Sorry, I haven't had much time to look at your PRs. I promise to get to them soon.

tristan957 (approved):

PR #261, opened 2024-10-06
Add meson as formatter

Totto16:
Add meson as formatter

Since meson 1.5.0 meson has a format subcommand, see [here](https://mesonbuild.com/Commands.html#format)

This is still WIP, as the formatter doesn't accept stdin via "-" atm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,6 @@ the [Material Design Icons](https://materialdesignicons.com/) project.
   introspection files:
   `"C_Cpp.default.configurationProvider": "mesonbuild.mesonbuild"`
 
-\* - requires an installation of [muon](https://muon.build).
-
 ## New Extension ID
 
 If you come from a previous installation, please make sure you are on the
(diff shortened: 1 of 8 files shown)


tristan957:
Super excited for this!

Totto16:
> Super excited for this!

Me too, but still figuring out, what the best way of doing this is.

I never managed to create an issue in meson, since some things don't work as expected. With muon you can use `"-"` as file and it reads from stdin, but with meson this doesn't work yet. I looked into the source code, but it's more complicated, since you have more options, can have multiple file and other more complicated things. It would be nice, If we would Be able to discuss this over there, and see what the best solution for that would be, as for now, in meson v1.5.0 I tried   `"/dev/stdin"` and `"/dev/fd/0"` But they worked only in a interactive terminal and not in the extension 😓 

You are also a contributor in meson, so maybe you have an idea how to do this in a way, that is already supported in v1.5.0 and we should create an issue there, so that it would be supported in a better way, than trugh a hack 🤔

tristan957:
Maybe that is an issue that I can get to in the future.

Totto16:
Since meson v1.7.0 released today, this now works, there are two version checks in the code atm, since only >=1.7.0 supports the needed feature, to use the stdin. I tried several ways to "polyfill" this behavior in >=1.5.0 < 1.7.0 but didn't succeed. 


Feel free to try this, it should work as expected.

@tristan957

PR #268, opened 2024-12-09
General improvments

Totto16:
This adds some general improvements and three bug fixes, I saw when implementing #261 

Bugs:
- The meson version wasn't detected correctly, since the `g` flag was used in the regex for the version number, so the resulting `match` array was `["1.6.0"]` instead of `["1.6.0","1","6","0"]`
- The version comparison was done manually in https://github.com/mesonbuild/vscode-meson/pull/268/commits/5e9c9e961e8195dcc7276686775a2ae8e8e4abfe#diff-9a3b5396b880a25586663b03f9bda56906914991b7b43b493283eca3ccc74fc3R28 and there we only checked `version[1] < 50`, which only worked correctly before `1.0.0` (thsi never showed, since after this check we additionally check, if the returned filename is an array or not, so this only cleans this up)
- The slicing of the version for meson was done incorrectly, `["1.6.0","1","6","0"].slice(1,.3)` results in `["1","6"]` and not in three components

Improvements:
- add Version class, that handles:
   - comparison, so that we don't make obvious errors and so that it's much simpler
   - toString function, so that it can be easily printed
   - check at runtime, if the input is correct, this can help detect feature  errors easier, if e.g. muon changes it's version string, that we parse, so that we get a nice and early error (this also helped my find bug 3)
   
   
Most of these things I "need" for #261 or at least it makes the life much more easy, (I print versions in some places, comparison is an improvement, that benefits also existing code) 

Feel free to suggest any changes 😄

### DIFF
--- a/src/formatters.ts
+++ b/src/formatters.ts
@@ -27,16 +27,16 @@ async function reloadFormatters(sourceRoot: string, context: vscode.ExtensionCon
   const name = extensionConfiguration("formatting").provider;
   const props = formatters[name];
 
-  const { tool, error } = await props.check();
-  if (error) {
-    getOutputChannel().appendLine(`Failed to enable formatter ${name}: ${error}`);
+  const checkResult = await props.check();
+  if (checkResult.isError()) {
+    getOutputChannel().appendLine(`Failed to enable formatter ${name}: ${checkResult.error}`);
     getOutputChannel().show(true);
     return disposables;
   }
 
   const sub = vscode.languages.registerDocumentFormattingEditProvider("meson", {
     async provideDocumentFormattingEdits(document: vscode.TextDocument): Promise<vscode.TextEdit[]> {
-      return await props.format(tool!, sourceRoot, document);
+      return await props.format(checkResult.tool, sourceRoot, document);
     },
   });
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,8 +1,60 @@
 import * as vscode from "vscode";
+import type { Version } from "./version";
 
 type Dict<T> = { [x: string]: T };
-export type Tool = { path: string; version: [number, number, number] };
-export type ToolCheckFunc = () => Promise<{ tool?: Tool; error?: string }>;
+
+export type Tool = { path: string; version: Version };
+
+export type ToolCheckSuccessResult = {
+  tool: Tool;
+  error?: undefined;
+};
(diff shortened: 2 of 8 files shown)


src/types.ts
```diff
@@ -1,8 +1,21 @@
...
+export type Tool = { path: string; version: Version };
+
+export interface ToolCheckSuccessResult {
+  tool: Tool;
+  error?: undefined;
+}
+export interface ToolCheckErrorResult {
+  tool?: undefined;
+  error: string;
+}
```
tristan957, line 15: why an interface vs a type?

Totto16: Just out of habit, normally my linter / the linter settings complain on object types, so I already write them this way, but if this codebase prefers `type xy = {...}` I'll chnage it

Totto16: I changed it to use `type`, the linter accepts it, and literally one line above also type is used xD

src/formatters.ts
```diff
@@ -27,16 +27,16 @@ async function reloadFormatters(sourceRoot: string, context: vscode.ExtensionCon
   const name = extensionConfiguration("formatting").provider;
   const props = formatters[name];
 
-  const { tool, error } = await props.check();
-  if (error) {
-    getOutputChannel().appendLine(`Failed to enable formatter ${name}: ${error}`);
+  const checkResult = await props.check();
+  if (checkHasError(checkResult)) {
```
tristan957, line 31: What do you think about making checkHasError a method on ToolCheckResult?

Totto16: That would work too, it would need to be a class, but it would look better 🤔

tristan957: That would be my preferred way of going about this if you could :smile:. Thanks for your work getting these 2 PRs over the hump!

Totto16: I implemented it as class now (in a new commit, I can squash it later), the only thing that is not that nice, is that some getter functions have runtime error checking, since it's not possible to narrow class members with the `xy is XY` constraint as before, but for the 2 places where it's used, this works fine and looks nicer

tristan957 (approved):

tristan957:
Looks good. Please squash at will, and I will merge!

Totto16:
Squashed it to 2 commits, which makes the most sense to me

PR #272, opened 2025-01-06
feat: support (Neo)Vim by coc.nvim

Freed-Wu:
Screenshot_20250103_155404

https://www.npmjs.com/package/coc-meson

Refer:
https://github.com/valentjn/vscode-ltex/blob/3d0cb8cd9b4d0dc8ef6c08a4f376767820678cf6/src/extension.ts#L8-L14

Difference is vscode-ltex/coc-ltex uses a customized python script
https://github.com/valentjn/vscode-ltex/blob/3d0cb8cd9b4d0dc8ef6c08a4f376767820678cf6/tools/patchForTarget.py
to patch code
we use c language macro preprocessor cpp

I can maintain the part of coc.nvim


Freed-Wu:
> I could be swayed by a better diff 

Do you mean something like https://github.com/wader/vscode-jq/pull/8/files ?

tristan957:
Yeah potentially.

PR #275, opened 2025-01-28
feat: add more sophisticated formatter selection

### LINKED ISSUE #274: Change the default formatter to the official one
`meson` has supported `format (fmt)` subcommand, since 1.5, and 1.6.1 released recently. 
So we could use the official one that do not need ex-dependency.

Totto16:
- allow the formatting provider to be set as `auto`, this means, select the best available one
- add internal priority for formatting providers

This adds an algorithm, that searches for the best available formatter. This results in the following behavior:
- if `formatting.provider` is set to `auto` we try to find a formatter in the following order `["meson", "muon"`], if one of them is found in this order, that one is used

- if `formatting.provider` is set to a specific formatter, it only tries that one and fails, if it didn't find that one

fixes #274

Let me know, me if this solution is acceptable. The problem with the previous code is, that it wasn't really designed to handle multiple formatting providers that well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,12 +192,13 @@
         },
         "mesonbuild.formatting.provider": {
           "type": "string",
-          "default": "muon",
+          "default": "auto",
           "enum": [
-            "muon",
-            "meson"
+            "auto",
+            "meson",
+            "muon"
           ],
-          "description": "Select which formatting provider to use"
+          "description": "Select which formatting provider to use. If set to 'auto', the best available one is picked"
         },
         "mesonbuild.formatting.muonConfig": {
           "type": "string",
--- a/src/formatters.ts
+++ b/src/formatters.ts
@@ -9,18 +9,71 @@ type FormatterFunc = (tool: Tool, root: string, document: vscode.TextDocument) =
 type FormatterDefinition = {
   format: FormatterFunc;
   check: ToolCheckFunc;
+  priority: number;
 };
 
+//NOTE: the highest priority number means it is tested first, the lowest is tested last
 const formatters: Record<FormattingProvider, FormatterDefinition> = {
   muon: {
     format: muon.format,
     check: muon.check,
+    priority: 0,
   },
   meson: {
     format: meson.format,
     check: meson.check,
+    priority: 1,
   },
 };
+
+type FormatterError = { provider: FormattingProvider; error: string };
+
+type BestTool = {
+  provider: FormattingProvider;
+  tool: Tool;
+};
+
+type BestFormatterResult = BestTool | FormatterError[];
+
+async function getBestAvailableFormatter(provider: FormattingProvider | "auto"): Promise<BestFormatterResult> {
+  if (provider !== "auto") {
+    const props = formatters[provider];
+
+    const checkResult = await props.check();
+    if (checkResult.isError()) {
+      return [{ provider, error: checkResult.error }];
+    }
+
+    return { provider, tool: checkResult.tool };
+  }
+
+  // sort the available providers by priority
+  const providerPriority: FormattingProvider[] = (Object.keys(formatters) as FormattingProvider[]).sort(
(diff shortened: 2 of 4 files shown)


tristan957:
> * allow the formatting provider to be set as null, this means, select the best available one
> 
>     * add internal priority for formatting providers

Instead of null, can we use "auto"?

> 
> This adds an algorithm, that searches for the best available formatter. This results in the following behavior:
> 
>     * if `formatting.provider` is set to `null` we try to find a formatter in the following order `["meson", "muon"`], if one of them is found in this order, that one is used
> 
>     * if `formatting.provider` is set to `"meson"` we try to find a formatter in the following order `["meson", "muon"`], even if we set the provider to be `"meson"` and meson is not found, muon is used

This behavior doesn't make much sense to me. If the user requests `meson`, they should get `meson` if available, else don't format.

>     * The same for muon, it is a priority if the user selected it, but meson might be used, if muon is not found
> 
> 
> The only problem I can see, if you want to disable the formatter, there is no way of doing that, but that was previously also not the case.
> 
> fixes #274
> 
> Let me know, me if this solution is acceptable. The problem with the previous code is, that it wasn't really designed to handle multiple formatting providers that well.

Let me know your thoughts!

Totto16:
> > * allow the formatting provider to be set as null, this means, select the best available one
> >   
> >   * add internal priority for formatting providers
> 
> Instead of null, can we use "auto"?

Yes that is also a good idea 👍🏼 


> > This adds an algorithm, that searches for the best available formatter. This results in the following behavior:
> > ```
> > * if `formatting.provider` is set to `null` we try to find a formatter in the following order `["meson", "muon"`], if one of them is found in this order, that one is used
> > 
> > * if `formatting.provider` is set to `"meson"` we try to find a formatter in the following order `["meson", "muon"`], even if we set the provider to be `"meson"` and meson is not found, muon is used
> > ```
> 
> This behavior doesn't make much sense to me. If the user requests `meson`, they should get `meson` if available, else don't format.

Ok, So I'll chnage it, to just do that on the `"auto"` value 👍🏼 

> > ```
> > * The same for muon, it is a priority if the user selected it, but meson might be used, if muon is not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The only problem I can see, if you want to disable the formatter, there is no way of doing that, but that was previously also not the case.
> > fixes #274
> > Let me know, me if this solution is acceptable. The problem with the previous code is, that it wasn't really designed to handle multiple formatting providers that well.
> 
> Let me know your thoughts!

tristan957:
Thanks for your contributions to the extension. I'll try and a get a release out after this PR merges.

Totto16:
I implemented the suggested chnages 😄

package.json
```diff
@@ -192,12 +192,13 @@
...
+          "default": "auto",
           "enum": [
-            "muon",
-            "meson"
+            "auto",
+            "meson",
+            "muon"
           ],
-          "description": "Select which formatting provider to use"
+          "description": "Select which formatting provider to use. If not set, the best available one is picked"
```
tristan957, line 201: I don't think this is accurate anymore

src/formatters.ts
```diff
@@ -9,39 +9,102 @@ type FormatterFunc = (tool: Tool, root: string, document: vscode.TextDocument) =
...
   format: FormatterFunc;
   check: ToolCheckFunc;
+  priority: number;
 };
 
 const formatters: Record<FormattingProvider, FormatterDefinition> = {
   muon: {
     format: muon.format,
     check: muon.check,
+    priority: 1, // higher means less priority
```
tristan957, line 19: Hmm, I would expect a higher priority value to mean higher priority, similar to `z-index`

tristan957:
If you squash your commits, I'll go ahead and merge this.

PR #278, opened 2025-03-14
muon lsp

annacrombie:
Adds muon as an available lsp.

I just finished adding a simple lsp frontend to muon.  It currently supports diagnostics, limited code completion (variables, functions, and methods), and goto definition.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,17 @@
     "configuration": {
       "title": "Meson build configuration",
       "properties": {
+        "mesonbuild.trace.server": {
+          "type": "string",
+          "default": "off",
+          "scope": "window",
+          "enum": [
+            "off",
+            "messages",
+            "verbose"
+          ],
+          "description": "Traces the communication between VS Code and the current Meson language server."
+        },
         "mesonbuild.selectRootDir": {
           "type": "boolean",
           "default": true,


src/lsp/muon.ts
```diff
@@ -0,0 +1,24 @@
+import * as vscode from "vscode";
+
+import { Executable } from "vscode-languageclient/node";
```
tristan957, line 3:
```suggestion
import type { Executable } from "vscode-languageclient/node";
```

package.json
```diff
@@ -98,6 +98,17 @@
...
+        "mesonbuild.trace.server": {
+          "type": "string",
+          "default": "off",
+          "scope": "window",
+          "enum": [
+            "off",
+            "messages",
+            "verbose"
+          ],
+          "description": "Traces the communication between VS Code and the current meson language server."
```
tristan957, line 110:
```suggestion
          "description": "Traces the communication between VS Code and the current Meson language server."
```

tristan957 (approved):
Very cool. Thanks for your work on this. I'll get lsp support into neovim for muon as well.

tristan957:
I can spin a new release after this PR is merged.

src/lsp/muon.ts
```diff
@@ -0,0 +1,24 @@
+import * as vscode from "vscode";
+
+import type { Executable } from "vscode-languageclient/node.js";
+import { LanguageServerClient } from "../lsp/index.js";
+
+export class MuonLanguageClient extends LanguageServerClient {
```
Jannik2099: Do you think supporting the auto download option is feasible?

You'd have to implement an artifact() method, see mesonlsp.ts

annacrombie: It is feasible, but I'm not interested in supporting this because I think that an editor auto-downloading tools like this is a bad idea.  I'd much rather let the user handle installation on their own, especially since muon is a general purpose meson alternative, so you want to be able to invoke it on the commandline as well.  And you get the most benefit if you are certain that the muon you are using is the same as the one providing the lsp.

PR #281, opened 2025-04-16
add lldb-dap support

Jannik2099:
`llvm-vs-code-extensions.lldb-dap` is the new official lldb extension. The API is largely the same

Also note that `Extension().isActive` has never been correct - it returns whether the extension has been started, not whether it is enabled in the current workspace. See https://stackoverflow.com/a/78703190


tristan957:
> In retrospect, I wonder if we should expose this as a configuration option

Sounds like a good idea. Can you add a subsequent PR?

tristan957 (approved):

PR #282, opened 2025-04-18
add coverage support

Jannik2099:
This adds coverage support from meson's xml report.
All data exposed by the coverage format is hooked up. This means full line-based coverage, and limited branch coverage (gcovr only tells if a branch site was missed completely or partially, but not which specific branches were missed).

[Attributable coverage](https://code.visualstudio.com/updates/v1_96#_attributable-coverage) as introduced in vscode 1.96 is not implemented, as meson currently does not have the necessary facilities.

The coverage API was finalized in vscode 1.88, so bump the minimum accordingly.

This is based on #260 since it makes use of the parallel test execution.


tristan957:
Can you rebase this? Thanks!

tristan957:
Why are we using an HTML parser to parse XML?

Jannik2099:
> Why are we using an HTML parser to parse XML?

domutils, domhandler and htmlparser2 are essentially one library split into multiple components. They implement both html and xml parsing, and are the most used html/xml parsing library in terms of weekly downloads.

tristan957:
Thank you for the explanation

Jannik2099:
ping?

also, it'd be nice if you could cut a release soon (with or without this PR) because of #281 . The debug functionality is currently useless out of the box.

PR #283, opened 2025-04-19
make debugger extension configurable

Jannik2099:


src/tests.ts
```diff
@@ -166,3 +153,54 @@ export async function testDebugHandler(
...
+        debugType = "lldb-dap";
+      } else {
+        vscode.window.showErrorMessage("No debugger extension found. Please install one and try again");
+      }
+
+      return debugType;
+    }
+    case "cpptools": {
+      if (!cppTools) {
+        vscode.window.showErrorMessage(`Chosen debugger extension ${chosenDebugger} is not available`);
```
libfud, line 179: Speaking as someone who isn't a TS dev, but wouldn't it make sense to bind this to a closure and then invoke that closure in each error case to minimize repetition?

libfud:
I don't see much of anything wrong with it off the top. I have mixed feelings on auto resolving to alphabetical order, but implementing some sort of preference based system instead of the alternative (specifiying explicitly, which is handled) is probably not worth the effort. The only other thing I can think of is checking if these extensions are and do what is desired, but I don't know enough about the extensions API to know how feasible that is.

src/tests.ts
Jannik2099: instead of three invocations of `showErrorMessage`, I'd end up with three invocations of `showErrorMessageWrapper`

realistically, this part gets updated once every couple years when someone publishes a new debugger extension, so I wouldn't overthink it.

Jannik2099:
> I have mixed feelings on auto resolving to alphabetical order

it's not so much alphabetic by intent as it is the existing order in which support for the extensions was added

> The only other thing I can think of is checking if these extensions are and do what is desired

this is not a free-form field. it's a choice of supported extensions. The `default` case is just a failsafe.

tristan957 (approved):

PR #290, opened 2025-07-29
Allow all path setting to have multiple elements

annacrombie:
## Motivation

To fully support `muon` as an alternative meson implementation.  This extension invokes $meson assuming it is meson, but muon has a different CLI with different subcommands and flags.  For compatibility, muon has a subcommand `meson` that translates any meson CLI syntax into the equivalent muon CLI syntax.  Thus something like `muon meson setup --prefix=foo build` would get translated to `muon setup -Dprefix=foo build`.  To make this seamless in vscode-meson, I think the best solution is to allow settings like `mesonPath` to support multiple values.  You can work around this today by creating a wrapper script, but that shouldn't be needed.  By supporting multiple values in mesonPath you also enable other use-cases such as specifying the python interpreter to use.

While I was at it, I also implemented substitution of ${workspaceDirectory} in all string settings values.  This is particularly handy if you bundle your build tools with your project:
```
	"mesonbuild.languageServerPath": "${workspaceFolder}/Tools/bin/muon",
	"mesonbuild.mesonPath": ["${workspaceFolder}/Tools/bin/muon", "meson"],
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,12 +184,34 @@
           "description": "Specify the list of options used for running benchmarks. --benchmark is implicit."
         },
         "mesonbuild.mesonPath": {
-          "type": "string",
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "minItems": 1
+            }
+          ],
           "default": "meson",
           "description": "Specify the meson executable to use"
         },
         "mesonbuild.muonPath": {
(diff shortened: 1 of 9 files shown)


Jannik2099 (changes requested):
Overall okay modulo the three comments.

I'm not a huge fan of having `string | string[]` in so many places but I guess it's a necessary evil.

annacrombie:
bump @Jannik2099

PR #324, opened 2026-06-03
Tweak 2 options in `package.json`

lazolv:
Both `mesonbuild.formatting.enabled` and `mesonbuild.linting.enabled` are ignored when a language server is found. For some reason `mesonbuild.formatting.enabled` is also false by default, I don't understand why. Recently I had a situation where I got so confused that I went on to investigate, thinking there was a bug in this extension's code, only to find out that it was just the option being false by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -218,7 +218,7 @@
         "mesonbuild.linting.enabled": {
           "type": "boolean",
           "default": true,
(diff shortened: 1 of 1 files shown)


lazolv:
I closed it because the branch name was misleading

